PR #70, opened 2016-09-20
MLIBZ-1317 Make forgotUsername(), verifyEmail(), and resetPassword() functions on the User class static

ThomasConner:
#### Description
`forgotUsername()`, `verifyEmail()`, and `resetPassword()` functions on the `User` class should be static and not require an active user to be executed.

#### Changes
- Made `forgotUsername()`, `verifyEmail()`, and `resetPassword()` functions on the `User` class.
- Added unit tests for the changes made on the `User` class.

### DIFF
--- a/src/entity/src/user.js
+++ b/src/entity/src/user.js
@@ -758,6 +758,8 @@ export class User {
   /**
    * Request an email to be sent to verify the users email.
    *
+   * @deprecated Use the static function verifyEmail().
+   *
    * @param {Object} [options={}] Options
    * @return {Promise<Object>} The response.
    */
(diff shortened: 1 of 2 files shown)


src/entity/src/user.js
```diff
@@ -785,18 +807,40 @@ export class User {
    * @return {Promise<Object>} The response.
    */
   async forgotUsername(options = {}) {
```
tejasranade: @ThomasConner  In the interest of avoiding confusion, should we deprecate this instance method? Seems like it's meaningless anyway to call `forgotUsername` when you already have a user instance. I have similar comments about a couple of other user verification APIs. We could start by adding deprecation notes and eventually taking them out. Thoughts?

ThomasConner: Yea I like that idea. Would you like me to add deprecation documentation for this release?

tejasranade: Yes, please. Please also verify that the docs refer to the correct API.

ThomasConner: Done

PR #74, opened 2016-11-07
MLIBZ-1368 Update errors to include name property

thediycoder:
#### Description
This pull request updates the errors in the SDK to include name properties set to the name of the error.

#### Changes
- Properly extend native error class
- Set name value to correct name of the error


thediycoder:
https://github.com/Kinvey/devcenter/pull/134 is related Dev Center change.

PR #77, opened 2016-11-15
MLIBZ-1452 Merge credentials and login data

thediycoder:
#### Description
This PR merges the data returned from a login request with the passed in credentials. This prevents from overriding properties that are used by developers.

#### Changes
- Merge data returned from login request with credentials.

### DIFF
--- a/src/entity/src/user.js
+++ b/src/entity/src/user.js
@@ -5,7 +5,7 @@ import { AuthType, RequestMethod, KinveyRequest, CacheRequest } from '../../requ
 import { KinveyError, NotFoundError, ActiveUserError } from '../../errors';
 import DataStore, { UserStore } from '../../datastore';
 import { Facebook, Google, LinkedIn, MobileIdentityConnect } from '../../identity';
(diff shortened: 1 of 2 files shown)


src/entity/src/user.js
```diff
@@ -273,10 +273,8 @@ export default class User {
     return request.execute()
       .then(response => response.data)
       .then((data) => {
-        if (credentials[socialIdentityAttribute]) {
-          data[socialIdentityAttribute] = credentials[socialIdentityAttribute];
-        }
-
+        data = assign(data, credentials);
+        delete data.password;
```
tejasranade: Where does the password get populated from? I'm just asking to make sure it's not persisted anywhere.

thediycoder: `credentials` could contain the password if the user uses `kinveyAuth` to login with. `delete data.password` removes the password property from the data object.

thediycoder: I committed and update to only merge the `_socialIdentity` property on `data`.

PR #80, opened 2016-12-02
MLIBZ-1531 Grouping does not work

thediycoder:
#### Description
This PR fixes a bug with grouping on the SDK. When trying to perform a group request, the SDK responds with the following error

```
{ 
  description: 'One of more identifier names in the request has an invalid format'
  debug: 'The entityName in the request starts with an _ (underscore)',
  code: 400 
}
```

#### Changes
- Send a group request with a `POST` and not a `GET`
- Set the aggregation on the request object
- Include the reduce function as a `reduce` property on the object representation of an aggregation
- Add unit test for grouping

### DIFF
--- a/src/aggregation.js
+++ b/src/aggregation.js
@@ -114,6 +114,7 @@ export default class Aggregation {
     const json = {
       key: this.key,
       initial: this.initial,
+      reduce: this.reduceFn,
       reduceFn: this.reduceFn,
       condition: this.query ? this.query.toJSON().filter : {},
       query: this.query ? this.query.toJSON() : null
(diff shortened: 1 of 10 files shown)


src/aggregation.js
```diff
@@ -114,6 +114,7 @@ export default class Aggregation {
     const json = {
       key: this.key,
       initial: this.initial,
+      reduce: this.reduceFn,
       reduceFn: this.reduceFn,
```
tejasranade: @thediycoder like we discussed, please add a separate ticket to remove the duplication here `reduce` and `reduceFn`.

thediycoder:
@tejasranade  Related DevCenter PR https://github.com/Kinvey/devcenter/pull/142

PR #84, opened 2016-12-09
Add X-Kinvey-Request-Id to errors

thomasconner:
#### Description
This PR adds `X-Kinvey-Request-Id` to errors. You can view the `X-Kinvey-Request-Id` related to a specific error by accessing the `kinveyRequestId` property on an error object.

```javascript
asyncNetworkCall()
  .catch(function(error) {
    // error.kinveyRequestId
  });
```

#### Changes
- Add `X-Kinvey-Request-Id` to error objects.
- Add unit tests to check if `X-Kinvey-Request-Id` is set properly on an error object.

### DIFF
--- a/src/errors/index.js
+++ b/src/errors/index.js
@@ -1,4 +1,5 @@
 import ActiveUserError from './src/activeUser';
+import BaseError from './src/base';
 import FeatureUnavailableError from './src/featureUnavailable';
 import IncompleteRequestBodyError from './src/incompleteRequestBody';
 import InsufficientCredentialsError from './src/insufficientCredentials';
@@ -22,6 +23,7 @@ import SyncError from './src/sync';
 // Export
 export {
   ActiveUserError,
+  BaseError,
   FeatureUnavailableError,
   IncompleteRequestBodyError,
   InsufficientCredentialsError,
--- a/src/errors/src/activeUser.js
+++ b/src/errors/src/activeUser.js
@@ -1,7 +1,7 @@
 import BaseError from './base';
 
 export default class ActiveUserError extends BaseError {
-  constructor(message = 'An active user already exists.', debug, code) {
-    super('ActiveUserError', message, debug, code);
+  constructor(message = 'An active user already exists.', debug, code, kinveyRequestId) {
+    super('ActiveUserError', message, debug, code, kinveyRequestId);
   }
 }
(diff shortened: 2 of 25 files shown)


test/unit/errors/activeUser.test.js
```diff
@@ -53,12 +53,25 @@ describe('ActiveUserError', () => {
...
+
+  describe('kinveyRequestId', () => {
+    it('should return undefined as the default value', () => {
+      const error = new ActiveUserError();
+      expect(error.kinveyRequestId).toEqual(undefined);
+    });
+
+    it('should return 1', () => {
+      const kinveyRequestId = 1;
+      const error = new ActiveUserError(null, null, null, kinveyRequestId);
```
tejasranade: I think a more practical way to test this is to try a request with a mock, and then test that the active user error gets thrown with the correct request ID. Our error test cases should actually test for the correct errors being thrown.

src/errors/index.js
```diff
@@ -22,6 +23,7 @@ import SyncError from './src/sync';
 // Export
 export {
   ActiveUserError,
+  BaseError,
```
tejasranade: What is the reason for `BaseError` to be in the exports?

src/errors/src/activeUser.js
```diff
@@ -1,7 +1,7 @@
 import BaseError from './base';
 
 export default class ActiveUserError extends BaseError {
-  constructor(message = 'An active user already exists.', debug, code) {
-    super('ActiveUserError', message, debug, code);
+  constructor(message = 'An active user already exists.', debug, code, kinveyRequestId) {
```
tejasranade: For a client-only error like the `ActiveUserError`, what should I (the developer) expect the value of `kinveyRequestId` to be? Is this documented?

thomasconner: They should expect the `kinveyRequestId` to be `undefined`. It will need to be documented.

test/unit/errors/activeUser.test.js
thomasconner: I updated the tests.

src/errors/index.js
thomasconner: Right now nothing uses it but could be in the future. Just good practice. I can remove it if necessary.

src/errors/src/activeUser.js
tejasranade: Makes sense. Please open a devcenter PR to go with this.

PR #85, opened 2016-12-12
Expose a global timeout get/set API

thomasconner:
#### Description
This PR adds a global timeout get/set API on the `Client` class. The global timeout can be changed to any integer value. The global timeout will be used by all request instances when a specific timeout value is not provided for a particular request instance. The default value is `60s`.

#### Changes
- Add global timeout get/set API to `Client` class.
- Add unit test to test global timeout for the `Client` class and `Request` class.


.env
```diff
@@ -2,7 +2,7 @@ BABEL_ENV=production
 KINVEY_ACL_ATTRIBUTE=_acl
 KINVEY_DATASTORE_NAMESPACE=appdata
 KINVEY_DEFAULT_API_VERSION=4
-KINVEY_DEFAULT_TIMEOUT=10000
+KINVEY_DEFAULT_TIMEOUT=60000
```
tejasranade: Reminder - this should go in docs and release notes.

test/unit/request/request.test.js
```diff
@@ -0,0 +1,29 @@
+import Request from '../../../src/request';
+import { Client } from '../../../src/client';
+import expect from 'expect';
+
+describe('Request', function() {
+  describe('constructor', function() {
```
tejasranade: Just out of curiosity, is there a way for us to mock request timeout, and test that to be the value `timeout`?

src/client.js
```diff
@@ -191,6 +199,25 @@ export class Client {
     this._appVersion = appVersion;
   }
 
+  get defaultTimeout() {
+    return this._defaultTimeout;
+  }
+
+  set defaultTimeout(timeout) {
```
tejasranade: Please add reference doc notes to these new getters and setters.

test/unit/request/request.test.js
thomasconner: I added a unit test.

PR #86, opened 2016-12-12
Improve Code Test Coverage

thomasconner:
#### Description
This PR improves the code test coverage by adding more unit tests.


test/unit/entity/acl.test.js
```diff
@@ -0,0 +1,319 @@
...
+
+    it('should be set to true', function() {
+      const acl = new Acl({ _acl: {} });
+      acl.globallyWritable = true;
+      expect(acl.isGloballyWritable()).toEqual(true);
+    });
+  });
+
+  describe('addReader()', function() {
+    it('should not add exiting reader', function() {
```
tejasranade: You mean "existing", I think.

test/unit/entity/acl.test.js
```diff
@@ -0,0 +1,319 @@
+import { Acl } from 'src/entity';
+import { randomString } from 'src/utils';
+import expect from 'expect';
+
+describe('Acl', function() {
```
tejasranade: For `ACL` and `Metadata`, the tests you've added are good. 
One test I suggest adding for each (if it doesn't already exist) is to mock an entity response that has `acl` and `kmd` values, and make sure they're deserialized to the right objects. This will ensure, for example, that we're reading the correct JSON keys for readers / writers / dates etc.

src/entity/src/metadata.js
```diff
@@ -34,12 +35,16 @@ export default class Metadata {
   }
 
   get emailVerification() {
-    return this.kmd.emailVerification.status;
+    if (isDefined(this.kmd.emailVerification)) {
```
thomasconner: Yes beacuse if `emailVerification` is `null` or `undefined` it will throw and error.

test/unit/entity/acl.test.js
thomasconner: Fixed

test/unit/entity/acl.test.js
thomasconner: I think this test would be a bit redundant. Since you can mock whatever you want that mock wouldn't be anymore correct then what I have. I do have unit tests on the user class that check `emailVerification` and other properties on the `acl` and `kmd` properties.

PR #95, opened 2017-01-12
MLIBZ-1586 Failed login produces a TypeError

thomasconner:
#### Description
Fixes a bug that produces a `TypeError` whenever an `InvalidCredentialsError` is encountered with a `NetworkRequest`.

#### Changes
- Fix code to be synchronous and not to use promise syntax.
- Add test case to catch issue

### DIFF
--- a/src/request/src/network.js
+++ b/src/request/src/network.js
@@ -394,76 +394,75 @@ export class KinveyRequest extends NetworkRequest {
       })
       .catch((error) => {
         if (error instanceof InvalidCredentialsError && retry === true) {
-          return LocalRequest.getActiveUser(this.client)
-            .then((activeUser) => {
-              if (!isDefined(activeUser)) {
-                throw error;
-              }
+          const activeUser = LocalRequest.getActiveUser(this.client);
 
-              const socialIdentities = activeUser._socialIdentity;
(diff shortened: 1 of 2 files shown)


src/request/src/network.js
```diff
@@ -394,76 +394,75 @@ export class KinveyRequest extends NetworkRequest {
...
-              const socialIdentities = activeUser._socialIdentity;
-              const sessionKey = Object.keys(socialIdentities)
-                .find(sessionKey => socialIdentities[sessionKey].identity === SocialIdentity.MobileIdentityConnect);
-              const session = socialIdentities[sessionKey];
-
-              if (isDefined(session)) {
-                // Refresh MIC Token
-                if (session.identity === SocialIdentity.MobileIdentityConnect) {
-                  const refreshMICRequest = new KinveyRequest({
+          const activeUser = LocalRequest.getActiveUser(this.client);
```
thomasconner: @vinaygahlawat This line is the main change that is needed. You will notice it was changed from expecting a `Promise` to be returned to just the `activeUser`.

thomasconner: Other changes are shown due to having to reformat the indentation of the code.

PR #96, opened 2017-01-16
MLIBZ-1614: User Discovery

thomasconner:
#### Description
This PR adds the ability to discover users using the SDK.

#### Changes
- Override the `find()` function in the `UserStore` class to use the user discover api.
- Add a static `find()` function to the `User` class that proxy calls to the `UserStore`.
- Default all options to an empty object on the `User` class. Fixes MLIBZ-1579.
- Add unit tests for user discovery.

### DIFF
--- a/src/datastore/src/userstore.js
+++ b/src/datastore/src/userstore.js
@@ -1,5 +1,6 @@
 import { AuthType, RequestMethod, KinveyRequest } from '../../request';
 import { KinveyError } from '../../errors';
+import { KinveyObservable, isDefined } from '../../utils';
 import NetworkStore from './networkstore';
 import Promise from 'es6-promise';
 import url from 'url';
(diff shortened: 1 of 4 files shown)


src/datastore/src/userstore.js
```diff
@@ -33,6 +35,54 @@ class UserStore extends NetworkStore {
...
+   * @param {Query} [query] Query used to filter entities.
+   * @param {Object} [options] Options
+   * @param {Boolean} [options.discover] Discover users.
+   * @return {Observable} Observable.
+   */
+  find(query, options = {}) {
+    // If `options.discover`, use
+    // [User Discovery](http://devcenter.kinvey.com/guides/users#lookup)
+    // instead of querying the user namespace directly.
+    if (options.discover === true) {
```
tejasranade: @thomasconner Why are we overloading the `user.find` method? On iOS and Xamarin, for example, we have a separate method called `User.lookup()`. Since the REST API used for lookup is different, I'd prefer if we separated the methods. 
Btw, what is `find()` on users supposed to do? If it's a supported operation, we need unit tests to cover the expected result.

src/entity/src/user.js
```diff
@@ -889,13 +889,25 @@ export default class User {
...
 
   /**
+   * Find users.
+   *
+   * @param {Query} [query] Query used to filter entities.
+   * @param {Object} [options] Options
+   * @param {boolean} [options.discover] Discover users.
+   * @return {Observable} Observable.
+   */
+  static find(query, options = {}) {
```
tejasranade: Same comment as about `find` vs `lookup`. Also a side note - it's probably easier for us to be prescriptive about the methods that devs should use - i.e. we either expose this on `UserStore` or `User`. Maintaining both seems redundant and unnecessary overhead.

thomasconner: The `UserStore` extends the `NetworkStore` which provides some implementation that would otherwise be duplicated in the `User` class. However, I think at this point we should just expose the `User` class that is used to proxy the calls to the `UserStore` class.

src/datastore/src/userstore.js
thomasconner: In 1.x the [lookup](https://github.com/Kinvey/js-sdk/blob/1.x/src/core/user.js#L566) was performed using `user.find`. I think it makes more sense to separate it out but this is why I did it this way.

As for what a find does on the user collection, I assume it performs a find like it would on any other collection. Again looking at 1.x you will see that it contains code to perform a regular find. https://github.com/Kinvey/js-sdk/blob/1.x/src/core/user.js#L581.

thomasconner: Since this a `super` call to the implementation for `find()` on the `NetworkStore` do we need separate unit tests for the `UserStore`? The only change would be that the `pathname` is overridden. We could need 1 unit test to check this.

src/entity/src/user.js
tejasranade: That makes sense, as long as we do it consistently for all `UserStore` calls, and also deprecate the currently documented ones until devs switch over.

src/datastore/src/userstore.js
tejasranade: If you agree that separating the two methods them makes sense, let's do it. I think it reduces confusion for the end user, and it's better to be consistent with other v3 SDKs as compared to JS v1.

The unit test on `find()` was to clearly define what the expected behavior is (if we say `find` is a supported method on the `UserStore`). I suspect there's at least some difference in how it behaves on users vs data, but I'm not sure either.

src/entity/src/user.js
```diff
@@ -889,13 +889,35 @@ export default class User {
   }
 
   /**
+   * Find users.
+   *
+   * @param {Query} [query] Query used to filter entities.
+   * @param {Object} [options] Options
+   * @return {Observable} Observable.
+   */
+  static find(query, options = {}) {
```
tejasranade: Let's remove this until we know what we're expecting.

PR #101, opened 2017-01-17
MLIBZ-1604: Update a user causes 401 on subsequent API requests

thomasconner:
#### Description
This PR updates the active user data if a `PUT` request is made to update the user on the backend.

#### Changes
- Update active user data with data returned from `PUT` request made on the active user to the backend.
- Add unit tests for updating a user.

### DIFF
--- a/src/datastore/src/userstore.js
+++ b/src/datastore/src/userstore.js
@@ -51,7 +51,7 @@ class UserStore extends NetworkStore {
     }
 
     if (!data._id) {
-      return Promise.ject(new KinveyError('User must have an _id.'));
+      return Promise.reject(new KinveyError('User must have an _id.'));
     }
 
     return super.update(data, options);
--- a/src/entity/src/user.js
+++ b/src/entity/src/user.js
@@ -691,18 +691,17 @@ export default class User {
   update(data, options) {
     data = assign(this.data, data);
     return store.update(data, options)
-      .then(() => {
-        this.data = data;
-        return this.isActive();
-      })
-      .then((isActive) => {
-        if (isActive) {
-          return CacheRequest.setActiveUser(this.client, this.data);
+      .then((data) => {
+        if (this.isActive() === true) {
+          return CacheRequest.setActiveUser(this.client, data);
         }
 
-        return this;
+        return data;
       })
-      .then(() => this);
+      .then((data) => {
+        this.data = data;
+        return this;
+      });
   }
 
   /**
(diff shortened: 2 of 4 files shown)


src/datastore/src/userstore.js
```diff
@@ -51,7 +51,7 @@ class UserStore extends NetworkStore {
     }
 
     if (!data._id) {
-      return Promise.ject(new KinveyError('User must have an _id.'));
+      return Promise.reject(new KinveyError('User must have an _id.'));
```
tejasranade: A test to cover this error handling case would be nice.

test/unit/entity/user.test.js
```diff
@@ -300,4 +301,62 @@ describe('User', function() {
...
+          const responseData = assign(requestData, { _kmd: { authtoken: randomString() } });
+
+          // Kinvey API response
+          nock(this.client.apiHostname, { encodedQueryParams: true })
+            .put(`${user.pathname}/${user._id}`, requestData)
+            .reply(200, responseData);
+
+          return user.update({ email: email })
+            .then(() => {
+              expect(user.data).toEqual(responseData);
```
tejasranade: Does the updated user object, in this case, contain the auth token? I'd assume it doesn't. In either case, can we add an assert for the correct behavior?

src/datastore/src/userstore.js
thomasconner: A [unit test](https://github.com/Kinvey/js-sdk/pull/101/files#diff-ea2fab4905bfc00b94d59bb6949998e2R306) has already been written to cover this case.

tejasranade: Oops 😃

test/unit/entity/user.test.js
thomasconner: This [line](https://github.com/Kinvey/js-sdk/pull/101/files#diff-ea2fab4905bfc00b94d59bb6949998e2R323) is where the `authtoken` is recreated. This [line](https://github.com/Kinvey/js-sdk/pull/101/files#diff-ea2fab4905bfc00b94d59bb6949998e2R333) verifies the active user contains the new data including the new `authtoken`.

src/entity/src/user.js
```diff
@@ -691,18 +691,17 @@ export default class User {
...
     return store.update(data, options)
-      .then(() => {
-        this.data = data;
-        return this.isActive();
-      })
-      .then((isActive) => {
-        if (isActive) {
-          return CacheRequest.setActiveUser(this.client, this.data);
+      .then((data) => {
+        if (this.isActive() === true) {
```
thomasconner: Yes. I will remove.

PR #102, opened 2017-02-01
MLIBZ 1602: Remove password (hashes) from locally stored User object

thomasconner:
#### Description
This PR removes sensitive data from locally stored user objects.

#### Changes
- Delete password from data returned from API calls for `login()` and `me()`
- Delete password from data stored for active user
- Update related unit tests


tejasranade:
Turns out, this issue was logged incorrectly. The password is not being locally stored at all. The reporter (Ivo) mistook the authToken for the credentials. 
Regardless, the change in the PR is an additional safety check and the unit test applies.

PR #103, opened 2017-02-01
MLIBZ-1576: Kinvey V3 add acl to individual entity error

thomasconner:
#### Description
This PR fixes errors when trying to use `Kinvey.Acl` and `Kinvey.Metadata` with invalid objects.

#### Changes
- Create an `_acl` property when a provided entity to `Kinvey.Acl` does not contain an `_acl` property.
- Create an `_kmd` property when a provided entity to `Kinvey.Metadata` does not contain an `_kmd` property.
- Don't default entities to an empty object for `Kinvey.Acl` and `Kinvey.Metadata`.
- Update/Add related tests for `Kinvey.Acl` and `Kinvey.Metadata`.

### DIFF
--- a/src/entity/src/acl.js
+++ b/src/entity/src/acl.js
@@ -1,6 +1,5 @@
 import { KinveyError } from 'src/errors';
 import { isDefined } from 'src/utils';
-import cloneDeep from 'lodash/cloneDeep';
 import isPlainObject from 'lodash/isPlainObject';
 import isArray from 'lodash/isArray';
 import assign from 'lodash/assign';
@@ -13,8 +12,8 @@ import assign from 'lodash/assign';
  * var acl = new Kinvey.Acl(entity);
  */
 export default class Acl {
-  constructor(entity = {}) {
-    if (!isPlainObject(entity)) {
+  constructor(entity) {
+    if (isPlainObject(entity) === false) {
       throw new KinveyError('entity argument must be an object');
     }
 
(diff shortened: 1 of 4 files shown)


src/entity/src/acl.js
```diff
@@ -24,7 +23,8 @@ export default class Acl {
      * @private
      * @type {Object}
      */
-    this.acl = cloneDeep(entity._acl);
+    entity._acl = entity._acl || {};
+    this.acl = entity._acl;
```
tejasranade: @thomasconner the `Acl` class seems to have an `acl` variable. Can we instead just have the properties directly on the instance?

src/entity/src/metadata.js
```diff
@@ -19,7 +17,7 @@ export default class Metadata {
      * @private
      * @type {Object}
      */
-    this.kmd = cloneDeep(entity[kmdAttribute] || {});
+    this.kmd = entity._kmd || {};
```
tejasranade: Same comment as above. Do we need the `kmd` property?

src/entity/src/acl.js
thomasconner: No reason. Lets use your suggestion instead.

src/entity/src/metadata.js
thomasconner: Same as above

thomasconner: Fixed

src/entity/src/acl.js
thomasconner: Fixed

PR #104, opened 2017-02-13
MLIBZ-1525: Expired Session Crash

thomasconner:
#### Description
This PR verifies that we are able to correctly refresh MIC sessions when they have expired.

#### Changes
- Don't retry failed refresh requests when we are in the middle of refreshing a MIC session. [af4ce5f](https://github.com/Kinvey/js-sdk/commit/af4ce5fef3495dff74d418e94b356be60211b73c)
- Add unit tests to verify we can refresh MIC sessions. [9949d11](https://github.com/Kinvey/js-sdk/commit/9949d11855101f10dc99820f9b29cb662b1e68b0).

### DIFF
--- a/src/request/src/network.js
+++ b/src/request/src/network.js
@@ -425,7 +425,7 @@ export class KinveyRequest extends NetworkRequest {
                 properties: this.properties
               });
 
-              return refreshMICRequest.execute()
+              return refreshMICRequest.execute(false, false)
                 .then(response => response.data)
                 .then((newSession) => {
                   // Login the user with the new mic session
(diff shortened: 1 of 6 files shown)


src/request/src/network.js
```diff
@@ -425,7 +425,7 @@ export class KinveyRequest extends NetworkRequest {
                 properties: this.properties
               });
 
-              return refreshMICRequest.execute()
+              return refreshMICRequest.execute(false, false)
```
tejasranade: @thomasconner What are the two parameters here?

src/request/src/response.js
```diff
@@ -158,9 +158,9 @@ export class KinveyResponse extends Response {
       return super.error;
     }
 
-    if (isDefined(name)) {
-      error.name = name;
-    }
+    // if (isDefined(name)) {
```
thomasconner: It is deliberate. I wanted to make sure it was ok before I removed the code. Maybe we should discuss in standup.

src/request/src/network.js
thomasconner: The first parameter is used to return the raw response rather then just the data from the response. The second parameter is used to retry the request in the event a `401` status code is encountered.

thomasconner: See https://github.com/Kinvey/js-sdk/blob/9949d11855101f10dc99820f9b29cb662b1e68b0/src/request/src/network.js#L364

tejasranade: Cool. Makes sense. Do we have any requests that default to `retry = true`?

PR #105, opened 2017-02-14
MLIBZ-1507: Error Objects

thomasconner:
#### Description
This PR adds missing error objects and cleans up code for returning an error from a response.

#### Changes
- Add missing Error Objects. [6fcf28a](https://github.com/Kinvey/js-sdk/commit/6fcf28a7343ad7c91a17dd85be2a77ad9d1a934f) [7804891](https://github.com/Kinvey/js-sdk/commit/7804891abe96cf9d6fa759bff9a6e4706b6cae95)
- Add unit tests for error objects. [916d02d](https://github.com/Kinvey/js-sdk/commit/916d02d83a93e8f480f575a2d0bfa43946cdb2c5) [447f105](https://github.com/Kinvey/js-sdk/commit/447f105809ccaf2420c0bda1776751018e2faf29)

### DIFF
--- a/src/request/src/response.js
+++ b/src/request/src/response.js
@@ -1,22 +1,34 @@
 import Headers from './headers';
 import assign from 'lodash/assign';
 import {
+  APIVersionNotAvailableError,
+  APIVersionNotImplementedError,
+  AppProblemError,
+  BadRequestError,
+  BLError,
+  CORSDisabledError,
+  DuplicateEndUsersError,
   FeatureUnavailableError,
   IncompleteRequestBodyError,
+  IndirectCollectionAccessDisallowedError,
   InsufficientCredentialsError,
   InvalidCredentialsError,
   InvalidIdentifierError,
   InvalidQuerySyntaxError,
   JSONParseError,
   KinveyError,
+  KinveyInternalErrorRetry,
+  KinveyInternalErrorStop,
   MissingQueryError,
   MissingRequestHeaderError,
   MissingRequestParameterError,
   NotFoundError,
   ParameterValueOutOfRangeError,
-  ServerError
+  ServerError,
+  StaleRequestError,
+  UserAlreadyExistsError,
+  WritesToCollectionDisallowedError
 } from 'src/errors';
-import { isDefined } from 'src/utils';
 
 /**
  * @private
@@ -72,7 +84,6 @@ export default class Response {
     }
 
     const data = this.data || {};
-    const name = data.name || data.error;
     const message = data.message || data.description;
     const debug = data.debug;
     const code = this.statusCode;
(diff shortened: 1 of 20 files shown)


tejasranade:
@thomasconner  One of the comments from the ticket is the following - 

> InsufficientCredentialsError and InvalidCredentialsError has name value of “InsufficientCredentials” and “InvalidCredentials" respectively instead of “InsufficientCredentialsError” and “InvalidCredentialsError” - we have manually fixed this in method "error" of object KinveyResponse by overwriting the value of var name with the correct value e.g. ....else if (name === 'InsufficientCredentials') { name = 'InsufficientCredentialsError';.......}

This PR doesn't seem to fix that. Is there a documentation gap? We need to make it clear what the expected error names are.

src/request/src/response.js
```diff
@@ -150,18 +183,20 @@ export class KinveyResponse extends Response {
...
     } else if (name === 'ServerError') {
       error = new ServerError(message, debug, code, kinveyRequestId);
+    } else if (name === 'StaleRequest') {
+      error = new StaleRequestError(message, debug, code, kinveyRequestId);
+    } else if (name === 'UserAlreadyExists') {
+      error = new UserAlreadyExistsError(message, debug, code, kinveyRequestId);
+    } else if (name === 'WritesToCollectionDisallowed') {
+      error = new WritesToCollectionDisallowedError(message, debug, code, kinveyRequestId);
     } else {
       return super.error;
```
tejasranade: @thomasconner what is `super.error` here? If `name` is not one of the errors we expect, what will happen?

thomasconner: It will return an appropriate error based on the status code or default to a `KinveyError`. See https://github.com/Kinvey/js-sdk/blob/137cb29c64a837efca953c8c8af3b48d8105a271/src/request/src/response.js#L82

src/request/src/response.js
```diff
@@ -150,18 +183,20 @@ export class KinveyResponse extends Response {
...
     }
 
-    // if (isDefined(name)) {
```
thomasconner: By removing this code we fix the issue with `error.name` being the incorrect name. For example `InvalidCredentials` and not `InvalidCredentialsError`

src/request/src/response.js
tejasranade: Ok cool, makes sense.

src/request/src/response.js
tejasranade: I believe the developer who reported the ticket is looking for the opposite i.e. `InvalidCredentialsError`. It's possible that either our docs are incorrect or we used `InvalidCredentialsError` as the name in a previous release. In any case, we should make sure it's clear to them what to expect.

thomasconner: Sorry I meant now it will be `InvalidCredentialsError` and not `InvalidCredentials` for `error.name`.

PR #106, opened 2017-02-15
MLIBZ-1676: File Upload Bug

thomasconner:
#### Description
This PR fixes a bug that would prevent a file from being uploaded to GCS.

#### Changes
- Handle file upload responses correctly.
- Fix unit tests to ensure all mocked API responses are consumed and promises are either fulfilled or rejected when expected.

### DIFF
--- a/src/datastore/src/filestore.js
+++ b/src/datastore/src/filestore.js
@@ -256,17 +256,17 @@ class FileStore extends NetworkStore {
           .then((statusCheckResponse) => {
             Log.debug('File upload status check response', statusCheckResponse);
 
-            // Upload the file
             if (statusCheckResponse.isSuccess() === false) {
-              if (statusCheckResponse.statusCode !== StatusCode.ResumeIncomplete) {
-                throw statusCheckResponse.error;
-              }
-
-              options.start = getStartIndex(statusCheckResponse.headers.get('range'), metadata.size);
-              return this.uploadToGCS(uploadUrl, headers, file, metadata, options);
+              throw statusCheckResponse.error;
             }
 
-            return file;
+            if (statusCheckResponse.statusCode !== 308) {
+              return file;
+            }
+
+            // Upload the file
+            options.start = getStartIndex(statusCheckResponse.headers.get('range'), metadata.size);
+            return this.uploadToGCS(uploadUrl, headers, file, metadata, options);
           })
           .then((file) => {
             data._data = file;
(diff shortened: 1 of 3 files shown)


src/datastore/src/filestore.js
```diff
@@ -306,38 +305,34 @@ class FileStore extends NetworkStore {
...
-            const backoff = Math.pow(2, options.count) + randomInt(1000, 1);
-
-            // Throw the error if we have excedded the max backoff
-            if (backoff >= options.maxBackoff) {
-              throw response.error;
-            }
-
-            Log.debug(`File upload will try again in ${backoff} seconds.`);
+        if (response.statusCode === 308) {
+          Log.debug('File upload was incomplete. Trying to upload the remaining protion of the file.');
```
tejasranade: Typo - `protion`. Also, can you make this message more descriptive? You should mention why it's incomplete (server responded with 308).

src/datastore/src/filestore.js
```diff
@@ -306,38 +305,34 @@ class FileStore extends NetworkStore {
...
+          options.start = getStartIndex(response.headers.get('range'), metadata.size);
+          return this.uploadToGCS(uploadUrl, headers, file, metadata, options);
+        } else if (response.statusCode >= 500 && response.statusCode < 600) {
+          Log.debug('File upload error.', response.statusCode);
```
tejasranade: @thomasconner Again, if `response` has more information than the `statusCode`, it would be helpful to put it in the debug message.

test/unit/datastore/filestore.test.js
```diff
@@ -120,15 +124,20 @@ describe('FileStore', function() {
...
+        .then((data) => {
+          expect(data).toIncludeKey('_data');
+          expect(nock.isDone()).toEqual(true);
+          return data;
+        });
+      return promise.should.be.fulfilled;
     });
 
-    it('should resume a file upload when a 308 status code is received', async function() {
+    it('should resume a file upload when a 308 status code is received', function() {
```
tejasranade: @thomasconner  just out of curiosity.. if files were failing to upload to GCS before, how did this test pass?

thomasconner: The test was passing because we had the necessary requests mocked and the result of whether the file was uploaded to GCS or not is still the file data from Kinvey. Since we still were getting back the file data even though the file was not being sent to GCS the test appeared to be passing. I have added a check to ensure that all mocked responses are sent. If a mocked response, such as the GCS upload response, is not sent it will fail now.

src/datastore/src/filestore.js
thomasconner: I will see if I can add more information

src/datastore/src/filestore.js
thomasconner: I will see if I can add more information

src/datastore/src/filestore.js
tejasranade: Yeah, in the very least (since this is a debug log), add the fact that it's a 308, and if there's anything meaningful in the response body of the 308.

PR #110, opened 2017-02-20
MLIBZ-1626: Add unit tests for user logout

thomasconner:
#### Description
This PR adds unit tests for user logout.

#### Changes
- Add unit tests for user logout.

### DIFF
--- a/test/unit/entity/user.test.js
+++ b/test/unit/entity/user.test.js
@@ -211,6 +211,30 @@ describe('User', function() {
     });
   });
 
+  describe('logout()', function() {
+    beforeEach(function() {
+      return TestUser.login(randomString(), randomString());
+    });
+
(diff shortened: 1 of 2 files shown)


test/unit/entity/user.test.js
```diff
@@ -211,6 +211,30 @@ describe('User', function() {
     });
   });
 
+  describe('logout()', function() {
```
tejasranade: @thomasconner the `logout()` tests should assert more that just the active user API. Please add some asserts for the cache being cleaned, sync queue emptied and checks for any other local storage items we remove (e.g. device tokens, SQL tables etc)

thomasconner: @tejasranade I have added some checks for these items.

PR #115, opened 2017-03-20
MLIBZ-1089: Sending a POST request with `??` causes the request to be sent as a GET with Backbone

thomasconner:
#### Description
Issue describe here https://bugs.jquery.com/ticket/8417. By specifying `dataType: json` when sending a `POST` request using jQuery ajax that contains `??` in the body, the request is transformed to a `GET` request.

#### Changes
- Remove `dataType: json`

### DIFF
--- a/src/shims/backbone/persistence/net.js
+++ b/src/shims/backbone/persistence/net.js
@@ -164,7 +164,6 @@ var BackboneAjax = {
     var xhr = options.xhr = Backbone.ajax({
       complete    : onComplete,
       data        : body,
-      dataType    : 'json',
       headers     : headers,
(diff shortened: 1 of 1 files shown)


tejasranade:
@thomasconner change looks fine, please add in the PR comments what tests you performed on this change.

thomasconner:
I tested the Backbone sample app with 3.x and found that 3.x did not have the issue at all. I then tested the same sample app using 1.x and reproduced the issue right away. After searching on the internet, I made the suggested change described here https://bugs.jquery.com/ticket/8417. After making the change, the issue went away.

PR #116, opened 2017-03-20
MLIBZ-1661: Share auth info between 1.x SDK and 3.x SDK

thomasconner:
#### Description
Customers would like to update to 3.x SDK slowly by migrating some pages to 3.x while other pages still use 1.x. To make this process easier we need to share auth info between each version of the SDK.

#### Changes
- Store auth info under 3.x SDK key.


thomasconner:
Tested this with the HTML5 Starter. I logged in using 1.x SDK and loaded the books collection. I then switched over to the 3.x SDK and loaded the books collection. Everything worked as expected and I wasn't required to relogin using 3.x SDK.

thomasconner:
I have thought this over and think it would be better if 3.x looked for the auth info stored by 1.x. I will submit a new PR with the change.

PR #118, opened 2017-03-27
Use tls: true as default when finding files

thomasconner:
#### Description
We should always default to using `https` urls by default when looking up files.

#### Changes
- use `tls: true` as default for finding files

### DIFF
--- a/src/datastore/src/filestore.js
+++ b/src/datastore/src/filestore.js
@@ -75,7 +75,11 @@ class FileStore extends NetworkStore {
    * });
    */
   find(query, options = {}) {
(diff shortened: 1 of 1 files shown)


thomasconner:
I was testing with the HTML5 starter app and noticed that when I loaded the files the `_downloadUrl` was using `http` and not `https` by default. In the file guide on the Dev Center we make the statement `The library always communicates with Google Cloud Storage using the https protocol. If you want to use http instead, the options argument in the methods described below allow for a tls: false flag.` This changes makes this statement true. We should have been using `tls: true` by default all along.

PR #120, opened 2017-03-30
Sync with query

thomasconner:
#### Description
This PR allows a user to push only entities in the sync table matching the provided query.

#### Changes
- Find entities that match the query provided before pushing entities to the backend.
- Add a unit test to verify only entities matching the query are pushed.


thomasconner:
@heyzooi @tejasranade Can you review this so we can merge if accepting for the pending release today?

thomasconner:
@heyzooi can you review when you have a chance today? We are trying to do a release today and would like to merge it in before that release. Thanks

heyzooi:
@thomasconner sure doing it right now!

PR #121, opened 2017-03-30
MLIBZ-1759: Query matches RegEx

thomasconner:
#### Description
This PR refactors `query.matches()` and adds unit tests.

#### Changes
- Refactor `query.matches()`
- Add unit tests for `query.matches()`

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -476,40 +476,38 @@ export default class Query {
    * @returns {Query} The query.
    */
   matches(field, regExp, options = {}) {
+    const flags = [];
+
     if (!isRegExp(regExp)) {
       regExp = new RegExp(regExp);
     }
 
-    if ((regExp.ignoreCase || options.ignoreCase) && options.ignoreCase !== false) {
-      throw new QueryError('ignoreCase flag is not supported.');
-    }
-
     if (regExp.source.indexOf('^') !== 0) {
-      throw new QueryError('regExp must have `^` at the beginning of the expression ' +
-        'to make it an anchored expression.');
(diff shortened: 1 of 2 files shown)


src/query.js
```diff
@@ -476,40 +476,38 @@ export default class Query {
...
     }
 
-    if ((regExp.ignoreCase || options.ignoreCase) && options.ignoreCase !== false) {
-      throw new QueryError('ignoreCase flag is not supported.');
-    }
-
     if (regExp.source.indexOf('^') !== 0) {
-      throw new QueryError('regExp must have `^` at the beginning of the expression ' +
-        'to make it an anchored expression.');
+      throw new QueryError('regExp must have \'^\' at the beginning of the expression'
```
tejasranade: It'd be nice for this error to say something about the fact that currently, only anchored expressions are supported on Kinvey.

src/query.js
```diff
@@ -476,40 +476,38 @@ export default class Query {
...
+        + ' to make it an anchored expression.');
     }
 
-    const flags = [];
+    if ((regExp.ignoreCase || options.ignoreCase) && options.ignoreCase !== false) {
```
tejasranade: @thomasconner if it is not supported, why is there a flag at all? Have we declared it anywhere in the docs or API reference? Was it documented in 1.x?

thomasconner: I still need to look into this. I didn't find it in the dev center but they linked the reference docs. The flag is there because you can make a reg ex case insensitive. `^/foo/i`

src/query.js
thomasconner: Ok

PR #126, opened 2017-05-16
MLIBZ-1802: Logs

thomasconner:
#### Description
This PR adds logging to the SDK.

#### Changes
- Log outgoing requests and incoming responses.
- Log active user activity.
- Log queries.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "hellojs": "1.14.1",
     "local-storage": "1.4.2",
     "lodash": "4.17.4",
-    "loglevel": "github:thomasconner/loglevel",
+    "loglevel": "github:pimterry/loglevel",
     "promise-queue": "2.2.3",
     "qs": "6.4.0",
     "request": "2.81.0",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "hellojs": "1.14.1",
     "local-storage": "1.4.2",
     "lodash": "4.17.4",
-    "loglevel": "1.4.1",
+    "loglevel": "github:thomasconner/loglevel",
     "promise-queue": "2.2.3",
     "qs": "6.4.0",
     "request": "2.81.0",
--- a/src/query.js
+++ b/src/query.js
@@ -1,5 +1,3 @@
-import { QueryError } from 'src/errors';
-import { nested, isDefined } from 'src/utils';
 import sift from 'sift';
 import assign from 'lodash/assign';
 import isArray from 'lodash/isArray';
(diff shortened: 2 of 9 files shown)


src/request/src/request.js
```diff
@@ -177,6 +180,7 @@ export default class Request {
 
   toPlainObject() {
     return {
+      id: this.id,
```
tejasranade: @thomasconner Is there any other value to the `id` outside of logging it here? Do we tie the request ID, for example, through the pipeline that the request goes through?

src/utils/src/log.js
```diff
@@ -18,13 +22,9 @@ Log.methodFactory = function methodFactory(methodName, logLevel, loggerName) {
     }
   };
 };
+log.setLevel(log.getLevel());
```
tejasranade: @thomasconner this looks odd. This should be a No-Op if getters and setters do what they're supposed to :)

package.json
```diff
@@ -40,7 +40,7 @@
     "hellojs": "1.14.1",
     "local-storage": "1.4.2",
     "lodash": "4.17.4",
-    "loglevel": "1.4.1",
+    "loglevel": "github:thomasconner/loglevel",
```
tejasranade: @thomasconner what were the changes you made to the package? Do you plan to keep your fork updated periodically from the master?

src/query.js
```diff
@@ -763,6 +768,8 @@ export default class Query {
     const json = this.toPlainObject();
     data = sift(json.filter, data);
 
+    Log.debug('After applying query filter to data', data);
```
tejasranade: @thomasconner Have you considered that this data could be quite large and flood the log? Would it make sense to truncate or just print the count instead, for example?

src/request/src/request.js
thomasconner: This `id` is generated by the SDK and used to match up log statements for requests. Multiple requests might be logged at the same time and without some kind of identifier it would be hard to match log statements that are related. We need to communicate that this is not related to the `X-Kinvey-Request-Id` header by adding the information to the troubleshooting guide.

thomasconner: This is the related [JIRA ticket](https://kinvey.atlassian.net/browse/MLIBZ-2025).

package.json
thomasconner: I submitted a [PR](https://github.com/pimterry/loglevel/pull/104) that was merged into the loglevel master branch. We can use the latest release of loglevel instead of a personally maintained version.

PR #128, opened 2017-05-17
MLIBZ-1840: Active User Bug

thomasconner:
#### Description
This PR fixes a bug with updating an active user. If multiple requests were sent concurrently to update the active user, some requests might fail because for a short time period the active user is `null`.

Fixes https://github.com/Kinvey/html5-sdk/issues/13.

#### Changes
- Update active user immediately so that other requests can use the new active user.
- Keep setting of active user in order using a queue.
- Add tests for this bug.

### DIFF
--- a/src/request/src/cache.js
+++ b/src/request/src/cache.js
@@ -1,11 +1,11 @@
 import Promise from 'es6-promise';
+import Queue from 'promise-queue';
 import UrlPattern from 'url-pattern';
 import url from 'url';
-import localStorage from 'local-storage';
 import cloneDeep from 'lodash/cloneDeep';
 
 import Client from 'src/client';
-import { KinveyError, NotFoundError } from 'src/errors';
+import { KinveyError } from 'src/errors';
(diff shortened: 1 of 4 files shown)


alexbainbridge:
Thanks @thomasconner !

alexbainbridge:
Hi @thomasconner . Is there an (even very approximate) ETA for when this fix will make it into a release. I see you have merged into the main codebase but without a release, we can't get to it via NPM I think. 

Thanks!

thomasconner:
Hey @alexbainbridge. Sorry for the delay. I am planning to do a release tomorrow which will contain the fix. I send an update in this PR with the release is available.

thomasconner:
@alexbainbridge We just released 3.5.1 for the JavaScript SDKs that contains the this fix.

alexbainbridge:
Thanks! Testing it now....  seems good! (but hard to test with 100% assurance for the non-happening of an intermittent error)

PR #129, opened 2017-05-22
MLIBZ-1766: Add missing protocol for custom hostnames

thomasconner:
#### Description
Add a valid protocol to `apiHostname` and `micHostname` if either does not contain one.

#### Changes
- Use a regex to test for `https://` or `http://`

#### Tests
- Add tests to set `apiHostname` and `micHostname` to hostnames contains valid and invalid protocols.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -42,15 +42,43 @@ export default class Client {
     }, options);
 
     if (options.apiHostname && isString(options.apiHostname)) {
-      const apiHostnameParsed = url.parse(options.apiHostname);
-      options.apiProtocol = apiHostnameParsed.protocol || 'https:';
-      options.apiHost = apiHostnameParsed.host;
+      let apiHostname = options.apiHostname;
+
+      if (/^https?:\/\//i.test(apiHostname) === false) {
+        apiHostname = `https://${apiHostname}`;
+      }
+
+      const apiHostnameParsed = url.parse(apiHostname);
+
+      /**
+       * @type {string}
+       */
+      this.apiProtocol = apiHostnameParsed.protocol;
+
+      /**
+       * @type {string}
+       */
+      this.apiHost = apiHostnameParsed.host;
     }
 
     if (options.micHostname && isString(options.micHostname)) {
-      const micHostnameParsed = url.parse(options.micHostname);
-      options.micProtocol = micHostnameParsed.protocol || 'https:';
-      options.micHost = micHostnameParsed.host;
+      let micHostname = options.micHostname;
+
+      if (/^https?:\/\//i.test(micHostname) === false) {
+        micHostname = `https://${micHostname}`;
+      }
+
+      const micHostnameParsed = url.parse(micHostname);
+
+      /**
+       * @type {string}
+       */
+      this.micProtocol = micHostnameParsed.protocol;
+
+      /**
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -42,15 +42,43 @@ export default class Client {
     }, options);
 
     if (options.apiHostname && isString(options.apiHostname)) {
-      const apiHostnameParsed = url.parse(options.apiHostname);
-      options.apiProtocol = apiHostnameParsed.protocol || 'https:';
-      options.apiHost = apiHostnameParsed.host;
+      let apiHostname = options.apiHostname;
+
+      if (/^https?:\/\//i.test(apiHostname) === false) {
```
tejasranade: @thomasconner there's a slight difference between this and the previous behavior. In the new case, if someone uses `http://` as the prefix, we will attach `https://` in front of it. This seems like it may confuse developers. Can we use the parsed url like before to figure out the `apiProtocol`?

test/client.test.js
```diff
@@ -13,18 +13,42 @@ describe('Client', () => {
...
       const apiHostname = 'https://mybaas.kinvey.com';
       const client = new Client({ apiHostname: apiHostname });
       expect(client.apiHostname).toEqual(apiHostname);
     });
 
-    it('should be able to provide custom micHostname', () => {
+    it('should be able to provide custom apiHostname with protocol http:', () => {
+      const apiHostname = 'http://mybaas.kinvey.com';
+      const client = new Client({ apiHostname: apiHostname });
+      expect(client.apiHostname).toEqual(apiHostname);
```
tejasranade: @thomasconner this test is potentially missing that the `apiProtocol` in this case will be `https://` because it was added by the SDK.

test/client.test.js
```diff
@@ -13,18 +13,42 @@ describe('Client', () => {
...
+    it('should be able to provide custom micHostname with protocol https:', () => {
       const micHostname = 'https://myauth.kinvey.com';
       const client = new Client({ micHostname: micHostname });
       expect(client.micHostname).toEqual(micHostname);
     });
 
+    it('should be able to provide custom micHostname with protocol http:', () => {
+      const micHostname = 'http://myauth.kinvey.com';
+      const client = new Client({ micHostname: micHostname });
+      expect(client.micHostname).toEqual(micHostname);
```
tejasranade: Same comment as the test above, applies here as well.

src/client.js
thomasconner: If someone uses `http://` the `if` statement will equal true. The regex `/^https?:\/\//i` checks for `http://` and `https://` strings. 

However, I am open to take it a step further and parsing the url before to determine the protocol being used.

test/client.test.js
thomasconner: See related answer above regarding `http://` and `https://` protocols.

test/client.test.js
thomasconner: See related answer above regarding `http://` and `https://` protocols.

src/client.js
tejasranade: Ah, good point. I didn't read the regex correctly. In that case, I'm ok with your current proposal. Protocols outside of http(s) are not really something we need to handle.

PR #130, opened 2017-05-30
Replace node modules with npm modules

thomasconner:


src/identity/src/mic.js
```diff
@@ -146,7 +133,7 @@ export class MobileIdentityConnect extends Identity {
               redirected = true;
               popup.removeAllListeners();
               popup.close();
-              resolve(url.parse(event.url, true).query.code);
+              resolve(qs.parse(event.url.split('?')[1]).code);
```
tejasranade: @thomasconner the `url.split` may not have 2 (or more) parts. This code needs to be more defensive to account for a url with no query.

src/identity/src/mic.js
```diff
@@ -159,7 +146,7 @@ export class MobileIdentityConnect extends Identity {
...
```
tejasranade: See above for urls without a query.

src/identity/src/mic.js
```diff
@@ -211,7 +198,7 @@ export class MobileIdentityConnect extends Identity {
       const location = response.headers.get('location');
 
       if (location) {
-        return url.parse(location, true).query.code;
+        return qs.parse(location.split('?')[1]).code;
```
tejasranade: See above for urls without a query.

src/utils/src/promise.js
```diff
@@ -0,0 +1,95 @@
...
+    return value;
+  }
+
+  return Promise.resolve(value);
+}
+
+/**
+ * @private
+ */
+export class Queue {
```
tejasranade: @thomasconner can we use a good promise queue package instead of maintaining this ourselves?

src/utils/src/url.js
```diff
@@ -0,0 +1,61 @@
...
+import isArray from 'lodash/isArray';
+import isPlainObject from 'lodash/isPlainObject';
+import isString from 'lodash/isString';
+import isEmpty from 'lodash/isEmpty';
+import { isDefined } from './object';
+
+/**
+ * @private
+ */
+function serialize(obj, options = {}, prefix) {
```
tejasranade: @thomasconner similar to promise-queue, can you instead use a third-party package?

package.json
```diff
@@ -19,61 +19,57 @@
...
   },
   "dependencies": {
-    "append-query": "2.0.1",
-    "core-js": "2.4.1",
+    "buffer": "5.0.6",
     "es6-error": "4.0.2",
     "es6-promise": "4.1.0",
+    "events": "1.1.1",
     "fast-memory-cache": "2.0.4",
-    "hellojs": "1.14.1",
```
tejasranade: @thomasconner please check the documentation implications of removing this package.

thomasconner: We will need to update the DevCenter to show how to connect Social Network accounts without using native APIs provided by the SDK.

src/utils/src/promise.js
thomasconner: I have looked at other ones but I would prefer to just use this one. The amount of code is small and easy to understand.

src/utils/src/url.js
thomasconner: No there is not a third-party package that doesn't depend on nodejs modules.

test/entity/user.test.js
```diff
@@ -962,7 +962,7 @@ describe('User', function() {
       const user = new User({ _id: randomString(), email: randomString() });
 
       nock(this.client.apiHostname)
-        .delete(`${user.pathname}/${user._id}`)
+        .delete(`${user.pathname}/${user._id}?hard=false`)
```
tejasranade: Ideally, for a parameter that's not required by the backend, we shouldn't send a value unless the user specifically provided it.

PR #131, opened 2017-05-30
Store active user with native storage

thomasconner:

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -4,7 +4,7 @@ import isNumber from 'lodash/isNumber';
 
 import { KinveyError } from 'src/errors';
 import { Log, isDefined } from 'src/utils';
-import { CacheRequest } from './request';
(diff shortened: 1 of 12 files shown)


src/client.js
```diff
@@ -153,11 +153,29 @@ export default class Client {
    * @param {string}    [options.appVersion]                               App Version
    * @return {Promise}                                                     A promise.
    */
-  static initialize(options) {
+  static initialize() {
```
tejasranade: @thomasconner this seems to go back to `init` from `initialize`. I understand that you're trying to revert to synchronous initialization, but this would be a significant breaking change.

thomasconner: It does go back to `init` from `initialize`. I disagree that this could cause a significant breaking change. It is much harder to go from synchronous to asynchronous. This update will be provided without introducing a breaking change.

PR #133, opened 2017-06-29
Covert to typescript

thomasconner:
#### Description
Update source to use typescript.


ivanstoyanov:
@thomasconner can you explain why we are doing this? Not challenging, just trying to understand.

PR #134, opened 2017-07-07
Fix URL formation on Windows and URL symbols in custom ID breaking cache storage

georgiwe:
#### Description
This PR should fix MLIBZ-1919 and MLIBZ-1677.

#### Changes
Parse path segments for cache requests manually and remove library from dependencies. Add library for joining URLs and ensure paths don't start with multiple slashes, because that broke the assertions in the tests due to discrepancies between mocked URLs and requested URLs.

#### Tests
Added new test for "." in _id. Old tests pass.

@thomasconner Other special characters seem to fail. If you want to release today, you might take a look. Otherwise I will work it out on Monday.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "request": "2.81.0",
     "rxjs": "5.2.0",
     "sift": "3.2.7",
-    "uid": "0.0.2"
+    "uid": "0.0.2",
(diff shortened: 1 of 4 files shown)


thomasconner:
Please see latest commits that allow all special characters.

PR #136, opened 2017-07-07
MLIBZ-1899: Fix Error Inheritance

thomasconner:
#### Description
This PR fixes error inheritances and removes the use of `es6-error`.

#### Changes
- Updated all errors to use the new inheritance model recommended on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
   "dependencies": {
     "append-query": "2.0.1",
     "core-js": "2.4.1",
-    "es6-error": "4.0.2",
     "es6-promise": "4.1.0",
     "fast-memory-cache": "2.0.4",
     "hellojs": "1.14.1",
--- a/src/errors/src/activeUser.js
+++ b/src/errors/src/activeUser.js
@@ -1,7 +1,14 @@
 import BaseError from './base';
 
-export default class ActiveUserError extends BaseError {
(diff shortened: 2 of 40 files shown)


tejasranade:
@thomasconner this PR needs some clarification. 
- What was the specific problem that the ticket reported? Was it a problem with one type of error or every type of error? Why weren't the unit tests catching it?
- Is there any impact on the API? e.g. reading properties of error objects, Typescript definitions etc. 
- I read the MDN recommendation. so I see where this change comes from. However, it seems like there's a lot of duplicate code in each error function (everything other than the message). Can this be parameterized?

thomasconner:
@tejasranade 
1) The problem was that the user would do `console.log(error)` which would not print out the `message` property but instead the `debug`, `code`, and `kinveyRequestId` properties. This has been reported several times by customers. This is fairly difficult to catch in unit tests since when you access `error.message` directly it is available and correct. If a customer did `console.log(error.message)` they would see the message but without reading the devcenter troubleshooting guide most would miss this. This problem occurred with all the errors.
2) There is no impact on the API. 
3) I was trying to prevent the duplicate code with the solution we had before but I think due to the way things were being inherited (babel transpilation to es5) caused things to get messed up. Errors are very touchy for some reason when it comes to JavaScript and therefore is why the duplicate code is necessary.

PR #137, opened 2017-07-12
Sdk on windows

georgiwe:
#### Description
This is related to MLIBZ-1958. Setting environment variables has different syntax on Windows so I removed the setting of BABEL_ENV variable from the build command since it is in the .env file anyways. Babel doesn't seem to have any settings depending on its environment though, so this seems redundant.

Also altered the test for special symbols in the _id field slightly, to only create one item with all special symbols we agreed upon with Tejas.

#### Changes
No changes to the SDK itself, only to the build command and one test, as described above.

### DIFF
--- a/src/request/src/cache.js
+++ b/src/request/src/cache.js
@@ -2,8 +2,6 @@ import Promise from 'es6-promise';
 import Queue from 'promise-queue';
 import url from 'url';
 import cloneDeep from 'lodash/cloneDeep';
-import identity from 'lodash/identity';
-import filter from 'lodash/filter';
 
 import Client from 'src/client';
 import { KinveyError } from 'src/errors';
@@ -70,9 +68,9 @@ export default class CacheRequest extends Request {
 
   set url(urlString) {
     super.url = urlString;
-    let pathname = global.decodeURIComponent(url.parse(urlString).pathname);
-    const urlParts = pathname.replace(/^\//g, '').split('/');
-    // "pathname" has the following form "/namespace/appKey/collection/id"
+    const pathname = global.decodeURIComponent(url.parse(urlString).pathname);
+    const urlParts = pathname.replace(/^\//, '').split('/');
+    // "pathname" has the following form: "/namespace/appKey/collection/id"
     this.appKey = urlParts[1];
     this.collection = urlParts[2];
     this.entityId = urlParts[3];
--- a/test/datastore/syncstore.test.js
+++ b/test/datastore/syncstore.test.js
@@ -386,17 +386,8 @@ describe('SyncStore', function() {
 
     it('should call update() for an entity that contains an _id with special characters', function() {
       const store = new SyncStore(collection);
-      const specialSymbols = ['.', '$', '~', '>', '<', '!', '@'];
-      const savePromises = specialSymbols.map(symbol => {
-        const id = `${randomString()}${symbol}${randomString()}`;
-        return store.save({ _id: id })
-          .then(resp => {
-            expect(resp).toExist();
-            expect(resp._id).toEqual(id);
-          });
-      });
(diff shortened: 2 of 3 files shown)


src/request/src/cache.js
```diff
@@ -70,9 +68,9 @@ export default class CacheRequest extends Request {
 
   set url(urlString) {
     super.url = urlString;
-    let pathname = global.decodeURIComponent(url.parse(urlString).pathname);
-    const urlParts = pathname.replace(/^\//g, '').split('/');
-    // "pathname" has the following form "/namespace/appKey/collection/id"
+    const pathname = global.decodeURIComponent(url.parse(urlString).pathname);
```
georgiwe: This went form global.escape() to global.decodeURIComponent, which do opposite things. I just wanted to point it out and make sure there isn't a specific reason to use escape previously. I also noticed url.parse() seems to do some arbitrary encoding - it encoded "<" and ">" but not unicode characters like "Ч". Couldn't find it in the docs. It's somewhat strange, but Tejas and I decided to leave it alone for the time being. Just thought I'd mention it.

tejasranade: @thomasconner - looks good to me, but leaving this open in case you want to take a look at Georgi's comment before it gets merged.

test/datastore/syncstore.test.js
```diff
@@ -386,17 +386,8 @@ describe('SyncStore', function() {
...
-        return store.save({ _id: id })
-          .then(resp => {
-            expect(resp).toExist();
-            expect(resp._id).toEqual(id);
-          });
-      });
-
-      return Promise.all(savePromises);
+      const id = '.$~<>!@+_#';
+      return store.save({ _id: id });
```
thomasconner: @georgiwe This change removes the expectations. Was this intended?

src/request/src/cache.js
thomasconner: @georgiwe I don't quite remember why `escape()` was used before. It could be possible that the `urlString` was not escaped before and therefore needed be but now it is by `url.parse()` although I am just speculating.

Regardless this change LGTM too.

src/request/src/cache.js
```diff
@@ -70,9 +68,9 @@ export default class CacheRequest extends Request {
...
+    const urlParts = pathname.replace(/^\//, '').split('/');
```
thomasconner: @georgiwe Why did the the __g__ flag get removed?

test/datastore/syncstore.test.js
georgiwe: It was, but now I'm thinking we better make sure the id is indeed the same. At the time I thought successfully saving the item was enough, but it is better to make sure. I will add the check for the id.

src/request/src/cache.js
georgiwe: Since the regexp is anchored with the symbol for start of line, it cannot occur anywhere else in the string and doesn't need to be global, I think. Should I change it back?

src/request/src/cache.js
georgiwe: We did observe some encoding done by parse(), but it wasn't in the docs so I'm confused. And double encoding (once in the core sdk and once in the NativeScript SDK) was the cause of a misrepresented auth token and 401 when trying to find() a file after uploading. So I think we should look into this and make a decision when how and where URLs are encoded and reach a "single point of truth" about it. Just not sure where :) I think somewhere in the core SDK...

src/request/src/cache.js
thomasconner: @georgiwe Ok makes sense. I have to admit I am not the best at regular expressions but I did use a tester to see the difference and I agree the `g` flag is not needed.

PR #138, opened 2017-07-20
[WIP] Refactor file upload

georgiwe:
***This PR features code which is merely a patch for another issue (the URL building). This is to keep the branch usable for the time being. This part of the PR (its first commit) is not meant to be merged so it's marked as WIP.***

#### Description
Part of the fix for MLIBZ-1965. NativeScript http module can (by design) only upload strings. A plugin will be used for files. Changes here were necessary since all of the file upload functionality was here, in the core SDK. This PR makes it possible to override the smallest necessary part of the upload process.

#### Changes
Refactored upload() method of FileStore class. Broke it up into smaller pieces in order to override one in a descendant class in NativeScript SDK. Namely, the makeUploadRequest() method.

#### Tests
Old tests pass (with the exception of 9 which fail due to URL changes in mocked and actual URLs, which is an entirely different problem due to the update of method of forming URLs). No new tests have been added since it's only refactoring.

### DIFF
--- a/src/datastore/src/filestore.js
+++ b/src/datastore/src/filestore.js
@@ -232,6 +232,90 @@ export default class FileStore extends NetworkStore {
   }
 
   /**
+   * Create the file on Kinvey
+   *
+   * @private
+   */
+  saveFileMetadata(options, requestBody) {
+    const isUpdate = isDefined(requestBody._id);
+    const reqMethod = isUpdate ? RequestMethod.PUT : RequestMethod.POST;
+    const reqUrl = url.format({
+      protocol: this.client.apiProtocol,
+      host: this.client.apiHost,
+      pathname: isUpdate ? `${this.pathname}/${requestBody._id}` : this.pathname
+    });
+
+    const reqOptions = {
+      method: reqMethod,
+      authType: AuthType.Default,
+      url: reqUrl,
+      properties: options.properties,
+      timeout: options.timeout,
+      body: requestBody,
+      client: this.client
+    };
+
+    const request = new KinveyRequest(reqOptions);
+    request.headers.set('X-Kinvey-Content-Type', requestBody.mimeType);
+    return request.execute();
+  }
+
+  /**
+   * @private
+   */
+  makeStatusCheckRequest(kinveySaveResponse, metadata, timeout) {
+    const headers = new Headers(kinveySaveResponse._requiredHeaders);
+    headers.set('content-type', metadata.mimeType);
+    headers.set('content-range', `bytes */${metadata.size}`);
+    const request = new NetworkRequest({
+      method: RequestMethod.PUT,
+      url: kinveySaveResponse._uploadURL,
+      timeout: timeout,
+      headers: headers
+    });
+    return request.execute();
+  }
+
+  /**
+   * @private
+   */
+  getUnexpectedStatusCheckResponsCode(response) {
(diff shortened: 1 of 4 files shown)


src/datastore/src/filestore.js
```diff
@@ -232,164 +232,239 @@ export default class FileStore extends NetworkStore {
...
-          });
+    return this.saveFileMetadata(options, metadata)
+      .then((response) => {
+        kinveySaveData = response.data;
+        return this.makeStatusCheckRequest(response.data, metadata, options.timeout);
+      })
+      .then((response) => {
+        Log.debug('File upload status check response', response);
+
+        if (response.isSuccess() === false) {
```
georgiwe: I haven't changed this logic, but shouldn't we add the backoff and retry here as well, for the status check request?

src/datastore/src/filestore.js
```diff
@@ -232,164 +232,239 @@ export default class FileStore extends NetworkStore {
...
+          return Promise.reject(response.error);
+        }
+
+        if (response.statusCode !== 308) {
+          // TODO: Here we should handle redirects according to location header, but this generally shouldn't happen
+          const err = this.getUnexpectedStatusCheckResponsCode(response);
+          return Promise.reject(err);
+        }
+
+        const uploadOptions = {
```
georgiwe: I had problems separating which methods care about which options, same with the responses, since objects were being mutated and I wanted to prevent this, so it's clear and no objects get changed by reference since it can get messy. I checked the docs and api reference but didn't see what options can be passed by the user, so I kept the ones which made the most sense - "timeout" and "maxBackoff". Any others I should pass along to the upload? The others seemed internal.

src/datastore/src/filestore.js
```diff
@@ -232,164 +232,239 @@ export default class FileStore extends NetworkStore {
...
+  handleReuploadCases(uploadResponse, fileSize, options) {
+    let backoff = 0;
+    const reuploadData = {
+      shouldResume: false,
+      shouldRetry: false
+    };
+
+    if (uploadResponse.isServerError()) { // should retry
+      Log.debug('File upload server error. Probably network congestion.', uploadResponse.statusCode, uploadResponse.data);
+      backoff = (2 ** options.count) + (1000 + randomInt(-100, 101)); // Calculate the exponential backoff
```
georgiwe: I rewroked this a bit, since the way it used to be was more random than it was exponential - you could get 1x1000, then 2x10, then 4x1000, 8x2. This version isn't all that mathematically proven, but I think it's a good balance between exponential with a jitter and simplicity :)

thomasconner: @georgiwe We had based the previous backoff calculation based on Google's [recommendation](https://cloud.google.com/storage/docs/exponential-backoff). I don't think we should change it.

georgiwe:
Hm... Well I personally believe small functions are better, I try to keep them under 20 lines usually. My guidelines during the refactoring were:

- separating each self-sufficient operation into a function - this way you give it a name and one doesn't need to read code, but rather just the name, in order to know what it does
- putting together these small pieces leading to less code in each complex operation

If you cannot follow the logic after the refactoring, it's not done well and we should definitely improve it. But I wish it isn't through making long functions, which do more than one thing, or logic is nested more than one level deep. We can talk it over after standup today, or at our meeting tomorrow. But if you insist, I will defer to your judgement, you can just make the changes as you see fit.

PR #139, opened 2017-07-24
MLIBZ-1967: Fix Storage API

thomasconner:
#### Description
Replaces all errors to use `KinveyError`. Export `Storage` class not as the default.

### DIFF
--- a/src/request/src/middleware/src/storage/index.js
+++ b/src/request/src/middleware/src/storage/index.js
@@ -31,7 +31,7 @@ export class Storage {
     return Promise.resolve()
       .then(() => MemoryAdapter.load(this.name))
       .then((adapter) => {
-        if (isDefined(adapter) === false) {
+        if (!isDefined(adapter)) {
           return Promise.reject(new KinveyError('Unable to load a storage adapter.'));
         }
 
@@ -65,7 +65,7 @@ export class Storage {
   }
 
   findById(collection, id) {
-    if (isString(id) === false) {
+    if (!isString(id)) {
       return Promise.reject(new KinveyError('id must be a string', id));
     }
 
@@ -87,7 +87,7 @@ export class Storage {
       }
 
       entities = entities.map((entity) => {
-        if (isDefined(entity._id) === false) {
+        if (!isDefined(entity._id)) {
           const kmd = entity._kmd || {};
           kmd.local = true;
           entity._kmd = kmd;
@@ -127,7 +127,7 @@ export class Storage {
 
   removeById(collection, id) {
     return queue.add(() => {
-      if (isString(id) === false) {
+      if (!isString(id)) {
         return Promise.reject(new KinveyError('id must be a string', id));
       }
 
--- a/src/request/src/middleware/src/storage/index.js
+++ b/src/request/src/middleware/src/storage/index.js
@@ -128,7 +128,7 @@ export class Storage {
   removeById(collection, id) {
     return queue.add(() => {
       if (isString(id) === false) {
-        return Promise.reject(new KinveyError('id must be a string'));
+        return Promise.reject(new KinveyError('id must be a string', id));
       }
 
       return this.loadAdapter()
--- a/src/request/src/middleware/src/storage/index.js
+++ b/src/request/src/middleware/src/storage/index.js
@@ -4,8 +4,8 @@ import isString from 'lodash/isString';
 import isArray from 'lodash/isArray';
 
 import { isDefined } from 'src/utils';
-import { NotFoundError } from 'src/errors';
-import MemoryAdapter from './src/memory';
+import { KinveyError, NotFoundError } from 'src/errors';
+import { MemoryAdapter } from './memory';
 
 Queue.configure(Promise);
 const queue = new Queue(1, Infinity);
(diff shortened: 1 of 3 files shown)


src/request/src/middleware/src/storage/index.js
```diff
@@ -66,7 +66,7 @@ export default class Storage {
 
   findById(collection, id) {
     if (isString(id) === false) {
-      return Promise.reject(new Error('id must be a string', id));
+      return Promise.reject(new KinveyError('id must be a string', id));
```
georgiwe: What is the "debug" property of the KinveyError object used for? Why add it here and but remove it on line 131?

src/request/src/middleware/src/storage/index.js
```diff
@@ -32,7 +32,7 @@ export default class Storage {
       .then(() => MemoryAdapter.load(this.name))
       .then((adapter) => {
         if (isDefined(adapter) === false) {
-          throw new Error('Unable to load a storage adapter.');
+          throw new KinveyError('Unable to load a storage adapter.');
```
georgiwe: Why "throw" instead of Promise.reject()? I didn't see this mentioned in the style guide, but I think it's worth deciding on one and using only it. I vote for Promise.reject(), since it's what one is trying to do when throwing inside a promise - reject the promise, not really throw an error, the rejection is essentially a side effect of the throwing. Seems semantically different to me. Also if you accidentally throw when you are not in a promise context, it would actually throw, so that's one more thing to think about.

src/request/src/middleware/src/storage/index.js
```diff
@@ -127,8 +127,8 @@ export default class Storage {
 
   removeById(collection, id) {
     return queue.add(() => {
-      if (!isString(id)) {
-        return Promise.reject(new Error('id must be a string', id));
+      if (isString(id) === false) {
```
georgiwe: The style guide said to [use shorthands for booleans](https://github.com/Kinvey/style-guides/blob/a6237b72b60bc14c8d5023a24ce9dcf3d8e78937/backend/es6.md#comparison--shortcuts) and I think it's neater.

(This is also the area of the removed second argument to KinveyError's constructor I mentioned in another comment)

src/request/src/middleware/src/storage/index.js
```diff
@@ -14,14 +14,14 @@ export {
   MemoryAdapter
 };
 
-export default class Storage {
+export class Storage {
```
georgiwe: Shouldn't this class be in a file of its own, instead of index.js? I'm a fan of one class per file and file names matching class names. And index.js to just be a barrel.

src/request/src/middleware/src/storage/index.js
```diff
@@ -112,7 +112,7 @@ export default class Storage {
   remove(collection, entities = []) {
     return Promise.all(entities.map((entity) => {
       if (typeof entity._id === 'undefined') {
```
georgiwe: Should probably use !isDefined(entity._id) here, or isUndefined().

src/request/src/middleware/src/storage/index.js
thomasconner: In this case, any consumer of the `Storage` class should only interface with the `Storage` class and not any of the storage adapters. By using `index.js` for the `Storage` class, import statements look like `import { Storage } form './storage';` If it was separated into a different file and then exported with an `index.js` file the effect would be the same. This removes that redundancy.

src/request/src/middleware/src/storage/index.js
thomasconner: I agree and this style guide didn't exist when I rewrote the JS SDK with ES6. I will replace all `throws` with `Promise.reject`.

src/request/src/middleware/src/storage/index.js
thomasconner: The `debug` property is used to provide extra helpful information on how to resolve or fix and error that may have occurred. It was really a carry over from version 1.x of the JS SDK and isn't as widely used in 3.x.

src/request/src/middleware/src/storage/index.js
thomasconner: You are right. Good catch

src/request/src/middleware/src/storage/index.js
thomasconner: Same as answer above. I will adopt the shorthand for booleans and start with this code.

thomasconner:
I have committed the changes requested. Please review when you have a moment.

src/request/src/middleware/src/storage/index.js
thomasconner: Fixed

src/request/src/middleware/src/storage/index.js
thomasconner: Fixed

src/request/src/middleware/src/storage/index.js
thomasconner: Fixed

src/request/src/middleware/src/storage/index.js
thomasconner: Fixed

src/request/src/middleware/src/storage/index.js
georgiwe: This is easily fixed by a good barrel file - export * from './storage'; inside index.js makes it a good barrel and you can import { Storage } from './storage'; as if it's defined inside index.js.

PR #142, opened 2017-08-03
Refactor

thomasconner:
#### Description
This refactor cleans up how the code is exported and the storage of the active user.

#### Changes
- Don't export any modules in `src/` as the default module.
- Import all modules correctly because of the change made above.
- Store the active user using the client into memory. Allow the storage of the active user to be overridden by any shims using the package.

#### Tests
- No tests were added but some tests were fixed to import modules from `src/` correctly.


tejasranade:
Replaced with separate PRs for active user changes (#143) and class export changes (#144)

PR #143, opened 2017-08-08
Changes to handle active user storage as a sync or async operation

tejasranade:
#### Description
- Nativescript persists the user synchronously while PhoneGap is async. This change allows the shims to handle active user persistence in the way that the respective platform supports.

#### Changes
- Store the active user using the client into memory. Allow the storage of the active user to be overridden by any shims using the package.

#### Tests
- None


src/request/src/network.js
```diff
@@ -387,52 +386,52 @@ export class KinveyRequest extends NetworkRequest {
...
-          const session = socialIdentities[sessionKey];
+          if (isDefined(activeUser)) {
+            const socialIdentity = isDefined(activeUser._socialIdentity) ? activeUser._socialIdentity : {};
+            const sessionKey = Object.keys(socialIdentity)
+              .find(sessionKey => socialIdentity[sessionKey].identity === 'kinveyAuth');
+            const oldSession = socialIdentity[sessionKey];
 
-          if (isDefined(session)) {
-            // Refresh MIC Token
-            if (session.identity === SocialIdentity.MobileIdentityConnect) {
```
tejasranade: @thomasconner Why was this check removed?

src/request/src/network.js
```diff
@@ -387,52 +386,52 @@ export class KinveyRequest extends NetworkRequest {
...
+              const request = new KinveyRequest({
                 method: RequestMethod.POST,
                 headers: {
                   'Content-Type': 'application/x-www-form-urlencoded'
                 },
                 authType: AuthType.App,
                 url: url.format({
-                  protocol: session.protocol || this.client.micProtocol,
-                  host: session.host || this.client.micHost,
+                  protocol: this.client.micProtocol,
```
tejasranade: Similar to the previous question - why remove the check for `session.protocol` and `session.host`? Seems like it might miss some case.

src/request/src/request.js
```diff
@@ -153,21 +153,22 @@ export default class Request {
...
-        response = new Response({
-          statusCode: response.statusCode,
-          headers: response.headers,
-          data: response.data
-        });
-      }
-
-      return response;
-    });
+    return this.rack.execute(this)
```
tejasranade: I believe we've discussed this before, but you may want to check if this change in the request is necessary.

test/entity/activeUser.test.js
```diff
@@ -1,20 +0,0 @@
-import expect from 'expect';
-import { ActiveUserHelper } from 'src/entity';
-import { randomString } from 'src/utils';
-
-describe('ActiveUser', function() {
```
tejasranade: @thomasconner can you add tests to replace these, or more generally, test the `setActiveUser` and `getActiveUser` APIs?

src/request/src/network.js
thomasconner: This was removed because the `find` a few lines up will only find any sessions matching `SocialIdentity.MobileIdentityConnect` (a.k.a `'kinveyAuth'`). This check was redundant.

test/entity/activeUser.test.js
thomasconner: Yes I will

src/client.js
```diff
@@ -1,15 +1,37 @@
...
+  get(key) {
+    try {
+      return JSON.parse(this.memory.get(key));
+    } catch (e) {
+      return null;
+    }
+  }
+
+  set(key, value) {
+    if (isDefined(value)) {
```
thomasconner: Ok

PR #145, opened 2017-08-16
Fix for URL / query encoding issues

tejasranade:
#### Description
We use a utility method to append queries in the SDK. This method was adapted from a node package "append-query". 
Recently we've seen an issue with queries in URLs not being properly encoded. Previously, the issue was mis-diagnosed to be in the {N} shim, but it turns out the `appendQuery` is the culprit. 
This PR fixes the issue.

#### Changes
- Minor change in `appendQuery` to fix a typo.
- Tests

#### Tests
- New tests to cover `appendQuery`.
- Manual test on {N} app to confirm the encoding fix.

### DIFF
--- a/src/utils/src/url.js
+++ b/src/utils/src/url.js
@@ -9,6 +9,7 @@ import { isDefined } from './object';
 
 /**
  * @private
+ * Adapted from https://github.com/lakenen/node-append-query, to fit a non-node runtime.
  */
 function serialize(obj, options = {}, prefix) {
   const str = [];
@@ -35,7 +36,7 @@ function serialize(obj, options = {}, prefix) {
     } else if (isPlainObject(val)) {
       query = serialize(val, options, key);
     } else {
-      query = options.encodeComponent === true ?
+      query = options.encodeComponents ?
         `${encodeURIComponent(key)}=${encodeURIComponent(val)}` :
         `${key}=${val}`;
     }
@@ -53,7 +54,7 @@ export function appendQuery(uri, query, options = {}) {
   const parts = url.parse(uri, true);
   const queryToAppend = isString(query) ? qs.parse(query) : query;
   const parsedQuery = assign({}, parts.query, queryToAppend);
-  options = assign({ encodeComponents: true, removeNull: true }, options);
+  options = assign({ encodeComponents: true, removeNull: false }, options);
   parts.query = null;
   const queryString = serialize(parsedQuery, options);
   parts.search = isDefined(queryString) && isEmpty(queryString) === false ? `?${queryString}` : null;
--- a/test/utils/url.test.js
+++ b/test/utils/url.test.js
@@ -1,0 +1,72 @@
+import { appendQuery } from 'src/utils';
+import nock from 'nock';
+import expect from 'expect';
+
+/**
+ * @private
+ * Adapted from https://github.com/lakenen/node-append-query
+ */
+describe('appendQuery', function(){
+    it('should append querystring to queryless url', function () {    
+        var result = appendQuery('http://example.com/foo', 'bar=baz&beep=boop')
+            , expected = 'http://example.com/foo?bar=baz&beep=boop'
+        expect(result).toEqual(expected);
+    });
+
+    it('should append querystring to url that already has a query', function () {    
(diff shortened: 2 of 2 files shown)


src/utils/src/url.js
```diff
@@ -9,6 +9,7 @@ import { isDefined } from './object';
 
 /**
  * @private
+ * Adapted from https://github.com/lakenen/node-append-query, to fit a non-node runtime.
  */
 function serialize(obj, options = {}, prefix) {
```
georgiwe: Why serialize manually, if qs library can do this? Is there something custom here? We could still delegate some of it to qs, no?

And isn't it better to only encode URLs in one place, if that's even possible?

test/utils/url.test.js
```diff
@@ -0,0 +1,72 @@
+import { appendQuery } from 'src/utils';
+import nock from 'nock';
```
georgiwe: Nock doesn't seem to be in use.

test/utils/url.test.js
```diff
@@ -0,0 +1,72 @@
...
+import expect from 'expect';
+
+/**
+ * @private
+ * Adapted from https://github.com/lakenen/node-append-query
+ */
+describe('appendQuery', function(){
+    it('should append querystring to queryless url', function () {    
+        var result = appendQuery('http://example.com/foo', 'bar=baz&beep=boop')
+            , expected = 'http://example.com/foo?bar=baz&beep=boop'
```
georgiwe: Pretty sure style doc has something to say about this :P

Noticed it while writing tests for LiveService today - why are we using es6 imports, but no other es6 code in tests? Arrow funcs won't let you do things like this.timeout(X), but that's rare and I think we should. I think all platforms support these features nowadays, but if we need to use babel for tests, I take this back :)

src/utils/src/url.js
tejasranade: Good question with the `qs` library. I was also concerned about using multiple utils to encode, but @thomasconner might know the reasoning behind it.

test/utils/url.test.js
tejasranade: Re: the style doc - I've tried to keep the tests as similar to the original as possible, so we can compare later if required. 

Re: es6 code, again a good question for @thomasconner !

thomasconner: Much of the test code is written using es6. We don't use arrow functions for the callback in the `it()` function because this breaks the scope that is provided from whom called the callback function. 

If in a `beforeEach()` I want to set a variable like the following
```javascript
beforeEach(function() {
   this.foo = 'bar';
});
```
And then I would like to access that variable in one of my `it()` functions I must use `function() {}` and not an arrow function.

This is one example why we use regular functions.

src/utils/src/url.js
thomasconner: I only copied the code from `node-append-query`. This is what it was doing but maybe it is possible we can use the `qs` library to serialize as well.

test/utils/url.test.js
georgiwe: Hm, I don't know if you've had a look yet, but in the tests for `LiveService` I did used a more... visible scope. I feel it's more clear and you can always tell where it starts and ends, while I personally have no idea when and where the "this" refers to the same object in different describes and hooks. And if I'm overwriting something. I'd rather use the global scope than `this` as a global scope, if that's what's going on. I saw in the `test/index.js` file a client is set to `this` and then it's used in other places as `this.client` and was confused. Is that the best practice for mocha or something?

thomasconner: This was the practice we used to share a common `client` across tests. `test/index.js` runs before any of the tests are run and is used as a setup script. At the time when the tests were being written, which was over a year ago, this was what made sense to do. Now that the SDK has matured some there might be other/better ways to manage the scope. Things like this are common in any codebase that spans over a great amount of time. Our priority has not been to rewrite the tests to since they have been working sufficiently up to this time.

tejasranade:
@georgiwe I tested queries and file uploads with a {N} app - found some other unrelated issues, but encoding seems to work correctly. I'm comfortable merging this if everything else looks ok.

PR #146, opened 2017-08-16
WIP: Live service user-to-user

georgiwe:
#### Description
MLIBZ-1856 Most of the functionality is there. Some details need to be ironed out, as well as some small refactoring. Integration with the live service for collections remains - there are tickets for subtasks.

#### Changes
Added classes implementing user-to-user communication.

#### Tests
Added some tests for LiveService class - more to come. Some old tests fail, but I believe this is due to some fixes from master missing here. Will make sure before PR to master.


src/live/src/user-to-user/stream.js
```diff
@@ -0,0 +1,245 @@
...
+  /**
+   * @private
+   * @param {string} substreamOwnerId
+   * @returns {Promise}
+   */
+  _requestPublishAccess(substreamOwnerId) {
+    return this._makeStreamRequest(`${substreamOwnerId}/publish`, RequestMethod.POST)
+      .then((resp) => {
+        if (resp.substreamChannelName) {
+          this.publishChannels[substreamOwnerId] = resp.substreamChannelName;
```
vinaygahlawat: When does this `publishChannels[id]` get cleared out?  Should this happen if a `StatusCode.Forbidden` is received?

src/live/src/user-to-user/stream.js
```diff
@@ -0,0 +1,245 @@
...
+   * @private
+   * @param {string} userId
+   * @returns {Promise}
+   */
+  _requestSubscribeAccess(userId) {
+    const requestBody = { deviceId: this.client.deviceId };
+
+    return this._makeStreamRequest(`${userId}/subscribe`, RequestMethod.POST, requestBody)
+      .then((response) => {
+        this.subscribeChannels[userId] = response.substreamChannelName;
```
vinaygahlawat: Same question for `subscribeChannels[id]` as above with `publishChannels[id]`.

src/live/src/user-to-user/stream.js
```diff
@@ -0,0 +1,245 @@
...
+        return promise;
+      });
+  }
+
+  /**
+   * @private
+   * @param {string} userId
+   * @returns {string} The name of the channel for that user
+   */
+  _buildChannelName(userId) {
```
vinaygahlawat: What is the reason for attempting to build the channel name on the client side?  The channel name should be received in a successful request for publish/subscribe access, and it is cached in `publishChannels[id]` and `subscribeChannels[id]` respectively.

src/live/src/user-to-user/stream.js
```diff
@@ -0,0 +1,245 @@
...
+   * @param {string} userId
+   * @param {Object} message
+   * @returns {Promise}
+   */
+  _publishWithAccessRequest(userId, message) {
+    return this._requestPublishAccess(userId)
+      .then(() => {
+        const channelName = this.publishChannels[userId] || this._buildChannelName(userId);
+        return this.liveService.publishToChannel(channelName, message);
+      });
```
vinaygahlawat: What happens if the initial `_requestPublishAccess` returns a `StatusCode.Forbidden` because the user does not have publish rights?  Does this need to be handled here?

georgiwe: The error is bubbled to the call to (for instance) `send()` so the developer can handle it. I don't think there's more we can do. We have called the `/publish` endpoint and if user had rights, this wouldn't happen.

If you look at `_publish()` of `Stream`, it checks if this is a substream the user has had rights to publish to (at some point since the app was started in the current session). If it hasn't, then the method you are referring to is called - requests access and then publishes, if user has access. The `_retriablePublish()` method is called only on subsequent `_publish()` calls, when the app has gotten approval from the `/publish` endpoint (meaning user has rights) at least once.

src/live/src/user-to-user/stream.js
georgiwe: Now it's only used as a fallback.

Initially, I wanted to directly publish and if I got a 403, only then call `/publish`. This way, you'd save one HTTP call in cases where you've still got permission from your last call to `/publish`, but cause of the caching 403 issue I talked about, this didn't work well, because it would cache that 403 for 30 seconds, despite you requesting and being granted access. If that had remained the flow, I'd need to build the channel name the first time. If it wasn't for that 403 issue, I would still prefer this method, but now that we know this can happen, the cautious way seems better, at least until we know why/when it happens.

src/live/src/user-to-user/stream.js
georgiwe: This method overwrites it every time a call to `/publish` is made (to request access). It is called when the first `_publish()` attempt is made, or when a 403 is received when publishing to PubNub, so it will get overwritten/updated when necessary.

src/live/src/user-to-user/stream.js
georgiwe: Answer is almost the same here - gets overwritten every time `_requestSubscribeAccess()` is called, which is every time `_subscribe()` is called, which is whenever any of the subscription methods are called - `listen()`, `follow()`.

Difference from `publishChannels` is that `_requestSubscribeAccess()` is called every time, since subscribing doesn't impact performance as much and one additional HTTP call isn't that big of a deal, especially since PubNub timetokens let you receive last 100 (or 200 was it) messages, even if you weren't subscribed to that particular channel, just the channel group.

Perhaps we should discuss this, since we are using a single channel group and it might happen that you subscribe and immediately get 100 messages. This might not be expected. Some manual work with the timetokens might be required, if we wish to avoid this.

src/request/src/network.js
```diff
@@ -179,6 +179,32 @@ export class KinveyRequest extends NetworkRequest {
...
+    }
+
+    let prm = KinveyRequest.execute(o);
+    if (dataOnly) {
+      prm = prm.then(r => r.data);
+    }
+    return prm;
+  }
+
+  static execute(options) {
```
thomasconner: I like this. 👍

src/live/src/user-to-user/live-service.js
```diff
@@ -0,0 +1,335 @@
...
+      this._pubnubListener.removeAllListeners(PubNubListener.unclassifiedEvents);
+    }
+  }
+
+  /**
+   * @param {string} channelName The name of the PubNub channel to publish to
+   * @param {Object} message The message to be published
+   * @returns {Promise}
+   */
+  publishToChannel(channelName, message) {
```
thomasconner: Does the public API require the end user to supply the `channelName` to publish a message?

src/live/src/user-to-user/live-service.js
```diff
@@ -0,0 +1,335 @@
...
+        return Promise.reject(resp);
+      });
+  }
+
+  /**
+   * Start listening for events for specified channel
+   * @param {string} channelName
+   * @param {LiveServiceReceiver} receiver
+   */
+  subscribeToChannel(channelName, receiver) {
```
thomasconner: Does the public API require the end user to supply the channelName to subscribe to a channel?

src/live/src/user-to-user/live-service.js
```diff
@@ -0,0 +1,335 @@
...
+      throw validationError;
+    }
+    this._subscribeToListener(channelName, receiver);
+  }
+
+  /**
+   * Stop listening for events for specified channel
+   * @param {string} channelName
+   */
+  unsubscribeFromChannel(channelName) {
```
thomasconner: Does the public API require the end user to supply the channelName to unsubscribe from a channel?

src/live/src/user-to-user/pubnub-listener.js
```diff
@@ -0,0 +1,56 @@
...
+ * @property {string} operation The operation the status is for. Example: PNSubscribeOperation
+ * @property {string[]} affectedChannels The channels affected in the operation
+ * @property {string[]} subscribedChannels All the current subscribed channels
+ * @property {string[]} affectedChannelGroups The channel groups affected in the operation
+ * @property {number} lastTimetoken The last timetoken used in the subscribe request
+ * @property {number} currentTimetoken The current timetoken fetched in the subscribe response, which is going to be used in the next request
+ */
+
+export class PubNubListener extends EventEmitter {
+  static statusPrefix = 'status:';
```
thomasconner: If this is a constant value then we should define it as a `const` outside of the class. By defining as a `static` variable on the class it allows the variable to possibly be changed outside of the class.

```javascript
const STATUS_PREFIX = 'status:';

export class PubNubListener extends EventEmitter {
  // ...
}
```

src/live/src/user-to-user/pubnub-listener.js
```diff
@@ -0,0 +1,56 @@
...
+  static unclassifiedEvents = 'pubNubEventsNotRouted';
```
thomasconner: If this is a constant value then we should define it as a `const` outside of the class. By defining as a `static` variable on the class it allows the variable to possibly be changed outside of the class.

```javascript
const UNCLASSIFIED_EVENTS = 'pubNubEventsNotRouted';

export class PubNubListener extends EventEmitter {
  // ...
}
```

src/live/src/user-to-user/stream.js
```diff
@@ -0,0 +1,249 @@
...
+   */
+  listen(receiver) {
+    const userId = this._client.activeUser._id;
+    return this._subscribe(userId, receiver);
+  }
+
+  /**
+   * Stops listening for messages sent to the active user
+   */
+  stopListening() {
```
thomasconner: Is this the same as calling `stream.unfollow('<userId>')`? The comment for the functions seem to indicate that they do the same thing.

src/live/src/user-to-user/stream.js
```diff
@@ -0,0 +1,249 @@
...
+    return this._publish(userId, message);
+  }
+
+  // Directed comm
+
+  /**
+   * Listens for messages sent to the active user
+   * @param {MessageReceiver} receiver
+   */
+  listen(receiver) {
```
thomasconner: Is this the same as calling `stream.follow('<userId>', () => {})`? The comment for the functions seem to indicate that they do the same thing.

src/kinvey.js
```diff
@@ -184,6 +185,8 @@ Kinvey.AuthorizationGrant = AuthorizationGrant;
 Kinvey.CustomEndpoint = CustomEndpoint;
 Kinvey.DataStore = DataStore;
 Kinvey.DataStoreType = DataStoreType;
+Kinvey.Stream = Stream;
+Kinvey.LiveService = LiveServiceFacade;
```
georgiwe: It's where user would register for live service (and get the pubnub auth keys and channel, etc.). It's also where they would attach the handlers for "global" pubnub events, such as disconnects.

I'm also thinking about attaching the `Stream` class to it, to make the connection more apparent. And if we are keeping the `StreamACL" class, it might get attached to it as well, or to the `Stream` class - haven't decided.

src/entity/src/user.js
```diff
@@ -498,7 +499,13 @@ export default class User {
       client: this.client
     });
 
-    return request.execute()
+    let prm = Promise.resolve();
+
+    if (LiveServiceFacade.isInitialized()) {
+      prm = LiveServiceFacade.unregister();
```
georgiwe: I thought about it, but I'm not sure.

Feels like hiding logic somehow. And calling a method, which just does nothing seems odd. I feel if you call it, it should either work, or return an error.

But I am flexible on this one :)

src/entity/src/user.js
```diff
@@ -718,7 +725,7 @@ export default class User {
     // Check if this is the active user
     if (this.isActive() === false) {
       return Promise.reject(
-        new KinveyError('This user must be the active user to register from real time.')
+        new KinveyError('This user must be the active user in order to register for real time.')
```
georgiwe: Umm this part of the `User` class I had left to discuss when we discuss or start merging the collection and user-to-user logic. I didn't want to break the collection before we decided what we're doing. Cause I think the register logic shouldn't even be in the `User` class, I think it belongs to the `LiveService` class.

src/live/src/user-to-user/live-service-facade.js
```diff
@@ -0,0 +1,58 @@
...
+import { PubNubListener } from './pubnub-listener';
+import { getLiveService } from './live-service';
+
+
+function _getLiveService() {
+  const client = Client.sharedInstance();
+  return getLiveService(client);
+}
+
+function register() {
```
georgiwe: I think even if we decide the public way to register would be user.register(), it should still be just a call to LiveService.register() - (as mentioned in a comment above) I feel the actual register logic for live service should be in the `LiveService` class.

src/live/src/user-to-user/live-service-facade.js
```diff
@@ -0,0 +1,58 @@
...
+  const liveService = _getLiveService();
+  const activeUser = User.getActiveUser();
+
+  return liveService.registerUser(activeUser)
+    .then((pubnubConfig) => {
+      const pubnubClient = new PubNub(pubnubConfig);
```
georgiwe: The user would only call `LiveServiceFacade.register()`. I separated the registration for live service and the initialization of the `LiveService` class in order to facilitate testing. If we instantiate these two inside `LiveService`, they cannot be mocked. Seems dependency injection is the way to go. I'm really curious about other options though.

src/live/src/user-to-user/live-service-facade.js
```diff
@@ -0,0 +1,58 @@
...
+      const listener = new PubNubListener();
+      liveService.initialize(pubnubClient, listener);
+    });
+}
+
+function unregister() {
```
georgiwe: Discussed it in the other comments. Lets take a vote on where it should be, I guess :D

src/live/src/user-to-user/live-service-facade.js
```diff
@@ -0,0 +1,58 @@
...
+  const liveService = _getLiveService();
+  return liveService.unregisterUser()
+    .then((resp) => {
+      liveService.uninitialize();
```
georgiwe: You mean not wait for the response from the HTTP call? Or you mean that we shouldn't unregister them? Why not (to either:))?

src/live/src/user-to-user/live-service-facade.js
```diff
@@ -0,0 +1,58 @@
...
+
+function offConnectionStatusUpdates(func) {
+  _getLiveService().offConnectionStatusUpdates(func);
+}
+
+function unsubscribeFromAll() {
+  _getLiveService().unsubscribeFromAll();
+}
+
+function isInitialized() {
```
georgiwe: One might need to know the current status of the `LiveService`. Perhaps after encountering an error during subscribe()/unsubscribe()... I think it's okay to expose it. Why not?

src/request/src/network.js
```diff
@@ -179,6 +179,32 @@ export class KinveyRequest extends NetworkRequest {
     this.trace = options.trace === true;
   }
 
+  static executeShort(options, client, dataOnly = false) {
```
georgiwe: My reasoning was that every time an HTTP request was made, the same 10 lines were written everywhere and then after every `execute()`, there was a `.then(r => r.data)`. I agree two such methods are a bit much. I just wanted to separate the "many assumptions" method from the "execute" method.

I agree the consumer should have the option to use anything and everything, but I think there should be a shorter version. I think in most cases (like 98%) you don't want the statusCode of the successful response, or want to use auth different from "session". Saving one line on a method you call 100 times is a big gain - improves readability, reduces chances of a typo.

Maybe we could merge `execute()` and `executeShort()` with some adjustments on the defaults?

src/live/src/user-to-user/live-service.js
georgiwe: Depends on what you are referring to as "public". In order to be used by `Stream`, `LiveService` should expose this as public, in terms of data hiding. But the `LiveServiceFacade` hides this public method of the class from the end user, so they cannot call it. That's the point of the facade. Perhaps it needs a better name, since now it features a bit of logic (the `registerUser()` to `initialize()`). I'm a little unsure about this part.

src/live/src/user-to-user/live-service.js
georgiwe: No, the answer to this is the same as the above. I should have read ahead. You can refer to the doc (which is once again public lol) on what the public (SDK consumer facing) API is.

src/live/src/user-to-user/live-service.js
georgiwe: Same as the last two :)

src/live/src/user-to-user/pubnub-listener.js
georgiwe: Yeah I thought about this one too. The point is the user of the class to be able to access it, so they can attach handlers to events in the way they would be emitted.

I'm considering two options here:
1) Add a method, which subscribes to both events by given channel name, which keeps the entire logic for naming the events emitted by `PubNubListener` inside it, which is awesome. But takes away some control. But it is worth exploring.
2) Adding a static getter, without a setter - effectively making the property readonly
3) Do it the way you mentioned, but export the const in order to use it outside

Taking votes here too :)

src/live/src/user-to-user/pubnub-listener.js
georgiwe: Yup, answer from above applies.

src/live/src/user-to-user/stream.js
georgiwe: Same as above - it is, if user id is the active user's id.

test/mocks/src/user.mock.js
```diff
@@ -110,6 +114,11 @@ export default class UserMock extends User {
     nock(this.client.apiHostname, { encodedQueryParams: true })
       .post(`${this.pathname}/_logout`)
       .reply(204);
+      
+    if (LiveServiceFacade.isInitialized()) {
```
tejasranade: Is the `user.logout()` behavior covered anywhere in tests? I may have missed it, but it'll be useful to test the state of the live service instances when a user is logged out.

src/kinvey.js
thomasconner: Ok makes sense. I think we should try to only expose `Kinvey.LiveService` and somehow merge `Stream` into this namespace. Seems like this is what you are suggesting.

src/entity/src/user.js
thomasconner: We do something similar for `User.logout()`. You can call this API even when there is not an active user and because the end result is the same. I would prefer to see where you call `LiveServiceFacade.unregister()` without worrying about calling `LiveServiceFacade.isInitialized()`.

src/live/src/user-to-user/live-service-facade.js
thomasconner: If we moved the register logic to `LiveServiceFacade` and expose it at `Kinvey.LiveService` then `LiveService.register()` would be exposed as a public API too along with `User.register()`. We should not expose both to prevent confusion.

src/live/src/user-to-user/live-service-facade.js
thomasconner: Sorry I was not very clear. I meant can we have `liveService.unregisterUser()` also uninitialize the live service internally to the function.

src/live/src/user-to-user/live-service-facade.js
thomasconner: To me it seems like we are requiring the end user to understand more about PubNub and how it works. They first have to register a user but why do I need to know if something is initialized? What is initialized? Did I register my user? Doesn't that initialize me?

These are the kind of questions I would have if I was a developer. Our goal is to make it as simple as possible so the end user doesn't need to understand the internals.

src/live/src/user-to-user/live-service.js
thomasconner: Ok I understand. It is hidden from the Public API provided to the end user.

src/live/src/user-to-user/live-service.js
thomasconner: Reply same as above

src/live/src/user-to-user/live-service.js
thomasconner: Reply same as above

src/live/src/user-to-user/stream.js
thomasconner: Ok makes sense

src/live/src/user-to-user/stream.js
thomasconner: Ok makes sense

src/kinvey.js
georgiwe: As per my latest changes (which I haven't pushed yet, but hope to do so before standup today), I've attached `StreamACL` to `Stream` and I will move `Stream` to `LiveServiceFacade`. Think this should work well indeed.

src/entity/src/user.js
georgiwe: I had mistakenly read that you meant to put the check inside `unregister()` of `LiveService`. I like it inside `unregister()` of `LiveServiceFacade`, will put it there.

src/live/src/user-to-user/live-service-facade.js
georgiwe: Definitely shouldn't expose both. If you guys prefer it on the `User` class, will move it there. But not the logic, just the call. I much rather the logic is kept inside `LiveService`. Does that sound good to you?

src/live/src/user-to-user/live-service-facade.js
georgiwe: I prefer to keep `LiveService`'s methods as simple as possible and organize more complex operations in `LiveServiceFacade`. Like `LiveServiceFacade.register()` takes care of registering the user and the initializing `LiveService`. `LiveServiceFacade.unregister()` unregisters the user and uninitializes `LiveService`. Does that make sense?

src/live/src/user-to-user/live-service-facade.js
georgiwe: I think from a user's standpoint, `LiveServiceFacade.register()` (and I am really not a fan of the "register" word here) is the initialization. The requirement to do an initialization (register for live service) will be known to them, this has been decided. And having the ability to check whether this has been done or not, at any point in time, seems valuable to me - if I am using the SDK, I would like to have that. If we have an initialization step (the "registration"), I think we should have a way to tell if it has been done. I don't think it tells them about PubNub, necessarily. It's an initialization step and they are going to know it exists, so might as well have a check for it.

Registration rant coming.. :)

This brings me back to the fact that I don't like the "register user for live service" part of the vocabulary. I see it more as a "login" of sorts, rather than a registration. An authentication step. You call an endpoint, provide credentials and get credentials for PubNub. It shouldn't be called "registration" for users who don't use the SDK (and the endpoint is /register-realtime - when we now call it "live service").

In the SDKs, the more I think about it, the more I convince myself this shouldn't be up to the user at all - we should do it implicitly (like we request permissions to publish/subscribe). The user doesn't care about "registering for live service" (whatever this would mean to them). Thy care about sending and receiving messages. The fact we choose to use PubNub and have an authentication step, should only be a concern for us, if user is using an SDK. I was very much in favour of this, before I "submerged" in the current state and totally forgot about it :) But you reminded me with this comment. @tejasranade If there's time, I'd like to do that. But I doubt there is, especially if it would need to be implemented in other SDKs too. But I want to have it out there, for the record :D

src/request/src/network.js
georgiwe: The name (and the very fact there are two so similar methods) is a work in progress, as is all of this PR. Just a place to discuss. Also, I wanted to test it out by using it for this feature.

I think there should be one shorthand method and I think it should be called `execute()` and have the defaults of the current `executeShort()`. Or others, based on what you guys think is the most used case.

I really don't want to be building my own URLs using a library each time I want to make a request, especially if half of the URL is the same for each request. This is why I don't think the current `execute()` is enough of a convenience and would prefer to keep something like the current `executeShort()`. Because as is, `execute` only removes the "new ....; .execute()` part. And there is much more that can be omitted most of the time.

test/mocks/src/user.mock.js
georgiwe: I covered the part about calling `unregistered()` if `LiveService` was initialized, upon logout. Tests also exist for the state of `LiveService` before and after initialization. Alos, there is only one `LiveService` instance.

PR #156, opened 2017-09-28
Replace UUID 

thomasconner:
#### Description
The [uuid](https://www.npmjs.com/package/uuid) package uses the NodeJS [crypto](https://nodejs.org/dist/latest-v6.x/docs/api/crypto.html) module to generate a uuid. On NativeScript, the crypto module is not available. This PR removes the uuid dependency and replaces it with a custom developed uuidv4 function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,7 @@
     "uid-safe": "2.1.5",
     "url": "0.11.0",
     "url-join": "2.0.2",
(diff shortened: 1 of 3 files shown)


thomasconner:
@georgiwe I left it in the format we had been using since we knew it worked. Maybe at a later date if we want to refactor things we can take a look at changing the format of the `deviceId`.

PR #158, opened 2017-10-12
Move HTML5 source to this repo

thomasconner:
As we discussed this will be a much bigger PR then the rest of the shims will involve. There was a lot of research being done with this move.


src/core/query/query.js
```diff
@@ -200,19 +190,19 @@ export default class Query {
     }
   }
 
-  /**
-   * Checks if the query is able to be run offline on the local cache.
-   * @return {Boolean} True if it is able to be run offline otherwise false.
-   */
   isSupportedOffline() {
-    let supported = true;
+    return Object.keys(this.filter).reduce((supported, key) => {
```
georgiwe: `findKey()` can take a predicate as a scond argument, think this would simplify things. This looks much more complicated than before, and again shouldn't be the focus of a PR which merges repos. Is something else going on here during the merge? The longer the merge takes, the longer we have to freeze code.

src/core/query/query.js
```diff
@@ -283,7 +273,7 @@ export default class Query {
    * @returns {Query} The query.
    */
   greaterThan(field, value) {
-    if (isNumber(value) === false && isString(value) === false) {
+    if (isNumber(value) === false && typeof value !== 'string') {
```
georgiwe: Same 3 lines in 4 places in this file - perhaps this should be a utility function or at least a validation function for this class.

src/core/utils/src/object.js
```diff
@@ -26,3 +30,18 @@ export function nested(obj, dotProperty, value) {
 
   return value || obj;
 }
+
+
+/**
+ * @private
+ */
+export function keyBy(array, iteratee) {
```
georgiwe: Are we trying to remove lodash as a dependency? I saw avoiding its functions and this is a function it has, being added.

scripts/postbuild.js
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
thomasconner: I have tried task runners in the past and where I do believe they have there place I am under the impression it is more overhead for use in our SDK then we need. I don't think we will have an explosion of scripts.

src/core/datastore/src/cachestore.js
```diff
@@ -4,17 +4,16 @@ import assign from 'lodash/assign';
 import keyBy from 'lodash/keyBy';
```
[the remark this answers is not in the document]
thomasconner: We will be removing the `../src/..` part with a future PR.

src/core/query/query.js
```diff
@@ -159,7 +149,7 @@ export default class Query {
    * @type {?number}
    */
   set limit(limit) {
-    if (isString(limit)) {
+    if (typeof limit === 'string') {
       limit = parseFloat(limit);
```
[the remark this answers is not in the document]
thomasconner: I think [line 166](https://github.com/Kinvey/js-sdk/pull/158/files/654755393c3ea8f8cad96a2c186631c5233d8273#diff-c2d18a2904fbbd8bc393a7dd3a414f8fL166) will be sufficient to verify if limit is a number or not.

src/core/query/query.js
thomasconner: I was setting up unit tests and I chose to use the Query and Aggregation unit tests as a way to figure out how things are setup. Part of further PRs will be to remove the use of [loadsh](https://lodash.com) as it is the main source as to why our bundles are somewhat big in size. Many of the lodash functions can be replaced by plain JavaScript functions.

With that being said, yes this probably shouldn't be part of this PR but we will do the this in the future so I rather leave it now.

src/core/query/query.js
thomasconner: Not sure I follow the comment "same 3 lines in 4 places in this file". There is some repeating that we can look at moving to a utility function in a future PR.

src/core/utils/src/object.js
thomasconner: Yes we are going to remove [lodash](https://lodash.com/) as a dependency in the near future. Some functions we might still need so we will add them to the appropriate place in the utils folder. I don't foresee this happening much because most of the functions we are using from [lodash](https://lodash.com/) are validation functions.

thomasconner:
Merging since @georgiwe said he was fine to go ahead and merge.

PR #159, opened 2017-10-18
MLIBZ-2011: Move PhoneGap SDK

thomasconner:
This moves the PhoneGap SDK source code into this repository.


thomasconner:
No I hadn't until you mentioned treeshaking but it looks like it would be very useful.

PR #169, opened 2017-11-23
Tnedyalkov/common tests

tsvetomir-nedyalkov:
#### Description
The tests can be run on the HTML5 and Nativescript shims. For example to run them on the HTML5 shim, are needed the following steps:
1. Clone the tnedyalkov/html5-tests branch of the HTML5 shim repo
2. Clone the tnedyalkov/commonTests branch of the core js-sdk repo
3. run 'npm install' in the HTML5 shim root folder
4. Copy the 'test' folder from the cloned core js-sdk repo to 'node_modules/kinvey-js-sdk/test' in the HTML5 shim
5. run 'node runner-config.js' in the HTML5 shim root


test/integration/crud-entity/offline-crud.test.js
```diff
@@ -0,0 +1,179 @@
...
+              .then(() => {
+                return syncStore.findById(entity._id).toPromise()
+              })
+              .then(() => {
+                return done(new Error(shouldNotBeCalledErrorMessage));
+              })
+              .catch((error) => {
+                expect(error.name).to.equal(notFoundErrorName);
+                return syncStore.count().toPromise()
+                  .then((count) => {
```
georgiwe: This `.then()` can be chained after the first `.catch`. But more importantly, why is it expected to have one entity offline? Shouldn't it be deleted?

test/integration/crud-entity/offline-crud.test.js
```diff
@@ -0,0 +1,179 @@
...
+                    expect(count).to.equal(1);
+                    done();
+                  })
+              })
+              .catch(done);
+          });
+
+          it.skip('findById() should remove entities that no longer exist on the backend from the cache', (done) => {
```
georgiwe: Is this intentionally skipped?

test/integration/setup.js
```diff
@@ -0,0 +1,10 @@
+before(() => {
+  collectionName = externalConfig.collectionName;
```
georgiwe: These are set globally? I think it's better to have one config object and keep things in there, if we must use global variables. That way we are not polluting the global scope with N variables, just the one.

test/integration/sync.test.js
```diff
@@ -0,0 +1,363 @@
...
+  let networkStore;
+  let syncStore;
+  let cacheStore;
+  let storeToTest;
+  const notFoundErrorName = 'NotFoundError';
+
+  //validates Push operation result for 1 created, 1 modified and 1 deleted locally items
+  const validatePushOperation = (result, createdItem, modifiedItem, deletedItem, expectedServerItemsCount) => {
+    return new Promise((resolve, reject) => {
+      expect(result.length).to.equal(3);
```
georgiwe: There is no synchronicity here - this doesn't need to be inside the Promise constructor. And returning `networkStore.find().toPromise()` is enough - no need for the Promise constructor at all.

test/integration/sync.test.js
```diff
@@ -0,0 +1,363 @@
...
+          expect(record.entity).to.deep.equal(record._id === createdItem._id ? createdItem : modifiedItem);
+        }
+        else {
+          expect(record.entity).to.not.exist;
+        }
+      })
+      networkStore.find().toPromise()
+        .then((result) => {
+          expect(result.length).to.equal(expectedServerItemsCount);
+          expect(_.find(result, (entity) => { return entity._id === deletedItem._id; })).to.not.exist;
```
georgiwe: This can be simplified to `_.find(result, e => e._id === deletedItem._id)`. Same for the lines below. `let createdOnServer` can be const.

test/integration/sync.test.js
```diff
@@ -0,0 +1,363 @@
...
+        })
+        .then((count) => {
+          expect(count).to.equal(0);
+          resolve();
+        }).catch(reject);
+    });
+  }
+  //validates Pull operation result
+  const validatePullOperation = (result, expectedItems, expectedPulledItemsCount) => {
+    return new Promise((resolve, reject) => {
```
georgiwe: Same comments as the above function - redundant Promise construction. And the lambdas.

test/integration/crud-entity/offline-crud.test.js
```diff
@@ -0,0 +1,179 @@
...
+      let networkStore;
+      let syncStore;
+      let cacheStore;
+      let storeToTest;
+      const dataStoreType = currentDataStoreType;
+      const entity1 = common.getSingleEntity(common.randomString());
+      let createdUserIds = [];
+
+      before((done) => {
+        common.cleanUpAppData(collectionName, createdUserIds)
```
georgiwe: I've seen this code one or two times before - perhaps it deserves its own function?

test/integration/users/users.tests.js
```diff
@@ -0,0 +1,614 @@
...
+      expect(user.data.username).to.exist;
+    }
+    if (shouldReturnPassword) {
+      expect(user.data.password).to.exist;
+    }
+    expect(user.isActive()).to.equal(true);
+    expect(user).to.deep.equal(Kinvey.User.getActiveUser());
+  }
+
+  const getMissingUsernameErrorMessage = () => {
```
georgiwe: Why is a function better than a string?

test/integration/common.js
```diff
@@ -0,0 +1,201 @@
...
+        .then(() => syncStore.pull())
+        .then(result => _.sortBy(common.deleteEntityMetadata(result), '_id'));
+    },
+    deleteUsers: (userIds) => {
+      return Promise.all(userIds.map(userId => {
+        return Kinvey.User.remove(userId, {
+          hard: true
+        })
+          .then(() => {
+            userIds.length = 0
```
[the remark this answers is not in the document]
tsvetomir-nedyalkov: As the function is used to delete user ids from the createdUserIds variable, which is initialized in every suite, the array should be emptied because we have to assure that there will be no attempt to delete the same user ids in a consequent call of deleteUsers.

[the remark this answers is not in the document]
georgiwe: I understand the array needs to be emptied, but I'm saying that it shouldn't be done in here. It is not apparent that the function will modify the array, it's an unexpected side effect.

You expect it to use it to delete entities, but not to remove them from the array. I feel whoever is calling the function should take care of emptying the array.

test/integration/common.js
```diff
@@ -0,0 +1,201 @@
...
+              expect(entityFromBackend).to.deep.equal(expectedEntity);
+            }
+            resolve();
+          }).catch(reject);
+      });
+    },
+    cleanUpCollectionData: (collectionName) => {
+      const networkStore = Kinvey.DataStore.collection(collectionName, Kinvey.DataStoreType.Network);
+      const syncStore = Kinvey.DataStore.collection(collectionName, Kinvey.DataStoreType.Sync);
+      return networkStore.find().toPromise()
```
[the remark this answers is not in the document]
tsvetomir-nedyalkov: networkStore.remove() does not work - the only option is by id or with a query

test/integration/crud-entity/common-crud.test.js
```diff
@@ -0,0 +1,1048 @@
...
+
+      before((done) => {
+        common.cleanUpAppData(collectionName, createdUserIds)
+          .then(() => {
+            return Kinvey.User.signup()
+          })
+          .then((user) => {
+            createdUserIds.push(user.data._id);
+            //store for setup
+            networkStore = Kinvey.DataStore.collection(collectionName, Kinvey.DataStoreType.Network);
```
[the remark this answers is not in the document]
tsvetomir-nedyalkov: As we discussed it is ok to stay as it is.

test/integration/crud-entity/offline-crud.test.js
tsvetomir-nedyalkov: One item is always created in the beforeEach and stays in the cache in order to be sure, that only the correct items are removed from it.

test/integration/crud-entity/offline-crud.test.js
tsvetomir-nedyalkov: Yes, it seems that there is a bug in the covered test case, but should be confirmed, so the tests is skipped for now.

test/integration/crud-entity/common-crud.test.js
```diff
@@ -0,0 +1,1048 @@
...
+                done(error);
+              }
+            });
+        });
+      });
+
+      describe('Querying', () => {
+        let entities = [];
+        const dataCount = 10;
+        const textFieldName = 'textField';
```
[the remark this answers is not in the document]
tsvetomir-nedyalkov: These variables are used when creating the entities...

test/integration/crud-entity/offline-crud.test.js
georgiwe: I see. 👍

test/integration/crud-entity/common-crud.test.js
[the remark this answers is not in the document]
georgiwe: This is what I mean: [link](https://github.com/Kinvey/js-sdk/blob/tnedyalkov/commonTests/test/integration/utilities.js#L24).

It should be something like this:
```javascript
const entity = {
  [textFieldName]: textValue || randomString()
};
```

test/integration/crud-entity/common-crud.test.js
```diff
@@ -0,0 +1,1048 @@
+runner.run(testFunc);
```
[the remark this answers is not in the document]
georgiwe: Test notifications.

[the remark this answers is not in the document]
georgiwe: Test 2

test/integration/crud-entity/common-crud.test.js
```diff
@@ -0,0 +1,1048 @@
...
+            .then(() => {
+              done();
+            }).catch((err) => {
+              done(err);
+            });
+        });
+
+        it('should update an existing entity', (done) => {
+          const entityToUpdate = {
+            _id: entity1._id,
```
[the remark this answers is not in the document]
tsvetomir-nedyalkov: Fixed the initially created entities to be used only by 'find and count operations' describe.

PR #170, opened 2017-11-28
MLIBZ-2201: Create Operation class

thomasconner:
I have added the `Operation` class and `OperationType` enum along with their associated unit tests. I have also added the some unit tests for the `DataStore` class.


packages/kinvey-datastore2/test/datastore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  it('should have type Sync', () => {
+    expect(DataStoreType.Sync).to.not.be.undefined;
+  });
+});
+
+describe('DataStore', () => {
+  describe('constructor', () => {
+    it('should throw and error if a collection is not defined', () => {
```
georgiwe: Typo. It's present in the other test with similar name too.

packages/kinvey-datastore2/test/datastore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      expect(() => new DataStore()).to.throw();
```
georgiwe: We should check the thrown error and make sure it's the error we expect. This way we avoid false positives when something else throws a different error.

packages/kinvey-datastore2/test/datastore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      const collectionName = 'test';
+      const datastore = new DataStore(collectionName);
+      expect(datastore.collection).to.equal(collectionName);
+    });
+  });
+
+  describe('save()', () => {
+    let datastore;
+
+    before(() => {
```
georgiwe: Why not just do `const dataStore = new DataStore()`, do we need the before?

packages/kinvey-datastore2/test/datastore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      datastore.save(entity, options);
+      expect(createSpy.called).to.be.true;
+      expect(createSpy.calledWith(entity, options)).to.be.true;
+    });
+  });
+
+  describe('create()', () => {
+    let datastore;
+
+    before(() => {
```
georgiwe: This seems common for all DataStore, perhaps put it in the `describe('DataStore', () => {...}`?

packages/kinvey-datastore2/test/datastore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      datastore = new DataStore('test');
+    });
+
+    it('should create an entity');
```
georgiwe: :)

packages/kinvey-datastore2/test/datastore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    expect(datastore).to.be.an.instanceof(DataStore);
+  });
+
+  it('should set the collection', () => {
+    const collectionName = 'test';
+    const datastore = collection(collectionName);
+    expect(datastore.collection).to.equal(collectionName);
+  });
+
+  it('should return a DataStore instance for DataStoreType.Cache');
```
georgiwe: How would we check for this?

packages/kinvey-datastore2/test/datastore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+  it('should return a DataStore instance', () => {
+    const datastore = collection('');
+    expect(datastore).to.be.an.instanceof(DataStore);
+  });
+
+  it('should set the collection', () => {
+    const collectionName = 'test';
+    const datastore = collection(collectionName);
+    expect(datastore.collection).to.equal(collectionName);
```
georgiwe: I think this is the only test we should have here - there is no logic in `collection()`, other than creating a DataStore instance.

Also having a `collection()` function, not tied to anything is weird - I think we should have it as part of a factory, or reconsider what makes the most sense for it. A function all by itself seems strange.

packages/kinvey-datastore2/test/operation.spec.js
```diff
@@ -0,0 +1,47 @@
+import { expect } from 'chai';
+import { Operation, OperationType } from '../src';
+
+describe('OperationType', () => {
+  it('should have type Create', () => {
+    expect(OperationType.Create).to.not.be.undefined;
```
georgiwe: Are we really writing unit tests for enums? I think they should at least feature the actual values - this way we are at least testing if any of them have changed unexpectedly. Having a defined value seems... inconclusive.

packages/kinvey-datastore2/test/operation.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+});
+
+describe('Operation', () => {
+  describe('constructor', () => {
+    it('should create an operation', () => {
+      const operation = new Operation();
+      expect(operation).to.be.an.instanceof(Operation);
```
georgiwe: What exactly are we testing here?

packages/kinvey-datastore2/src/operation.js
```diff
@@ -0,0 +1,14 @@
+export const OperationType = {
+  Create: 'Create'
+};
+Object.freeze(OperationType);
+
+export class Operation {
+  constructor(config = {}) {
```
georgiwe: I don't like this way of constructing an Operation instance. If we are doing OOD, we should stick to OO. I don't think Java or C# would use this manner of constructing objects.

packages/kinvey-datastore2/src/datastore.js
```diff
@@ -1,12 +1,35 @@
...
+    this.collection = collection;
+  }
+
   save(entity, options) {
     return this.create(entity, options);
   }
 
   create(entity, options) {
-    throw new Error('Method of DataStore not implemented');
+    const operation = this._buildOperationObject(OperationType.Create, null, entity);
```
georgiwe: I was really looking forward to seeing progress on this, rather than the tests, so I know how to proceed.

packages/kinvey-datastore2/src/operation.js
thomasconner: My only concern was the arguments to the constructor would keep growing. I am fine changing it back.

packages/kinvey-datastore2/test/datastore.spec.js
thomasconner: I was trying to stick to Mocha conventions for setup and teardown.

packages/kinvey-datastore2/test/datastore.spec.js
thomasconner: The problem is then in our `constructor` tests we already have a property called datastore that needs to be created for each test which would conflict.

packages/kinvey-datastore2/test/datastore.spec.js
thomasconner: Right now `collection()` only returns a DataStore instance but we will be adding more functionality to it in future PRs. This test will change accordingly.

It isn't all the strange in JavaScript to have a function by itself.

packages/kinvey-datastore2/test/datastore.spec.js
thomasconner: We would check that the DataStore instance had the correct processor.

packages/kinvey-datastore2/test/operation.spec.js
thomasconner: This unit test should check the value of the enum rather then if is is just defined.

packages/kinvey-datastore2/test/datastore.spec.js
georgiwe: It's odd calling collection() and getting a DataStore instance. For internal use, we should call it something else. And I feel `dataStoreFactory.create()` is more approprate. And when we expose it publicly, we can call it collection. I dunno, perhaps we can decide better later on.

packages/kinvey-datastore2/test/operation.spec.js
thomasconner: This was testing that you could create an Operation instance without passing any arguments. Not sure if it was necessary but it can't hurt to test. More tests are not a bad thing.

packages/kinvey-datastore2/src/operation.js
georgiwe: I know, 4+ constructor arguments feels odd. But I'd rather solve this differently. Perhaps the factory can instantiate an operation object and assign its properties the values it has been passed. Or we can branch off different Operation descendants. I've thought about it, but didn't feel pressing so I haven't done much in that direction.

packages/kinvey-datastore2/test/datastore.spec.js
georgiwe: Okay, I'm not too familiar with those. If that's the way the say we should do it, I'm fine with that.

packages/kinvey-datastore2/test/operation.spec.js
georgiwe: But they should bring value. Having more code to maintain comes at a price.

packages/kinvey-datastore2/test/datastore.spec.js
georgiwe: We can rename it. Or put it in a better scope. I vote for the former.

PR #171, opened 2017-11-28
MLIBZ-1999: Mono Repo

thomasconner:
This moves all the code from the individual repositories that make of the JavaScript SDKs into this repository as a mono repo format.


georgiwe:
I have a concern about the release procedure. To me, it seemed rather complicated in general. But the two parts I can point out as leaving the biggest impression are

- Travis taking actions on push of tags, regardless of the tags (this is my understanding of how it works) - maybe that's not a problem, but also adding a couple of validations shouldn't be too difficult either
- the bash file in each repo, which seems pretty important - I'm curious does that have to be committed in source control? isn't there a way to run a few lines of bash from Travis or someplace else?

These are the two things that stick out in my mind, but there was a lot more involved (understandably so) in the procedures that Travis runs. I think it's worth it going over those again.

thomasconner:
1) This one doesn't make sense to me. Why would a tag be created if it isn't suppose to be release?

2) Please take a look at https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps. This is the way travis recommends using a custom shell script for complex build steps. We need to push the `dist` created by travis back to the respective git repos in order for customers who use bower to be able to install the latest release. This custom shell script uses an access token that is encrypted for pushing the code back to the git repo. Am I missing something that causes this to be insecure?

PR #172, opened 2017-11-30
MLIBZ-2213 Subscribe to user's personal collection channel

georgiwe:
This fixes MLIBZ-2213.

#### Description
Fixes the lack of updates for entities, for which the user has inclusive permissions to view.

#### Changes
Added a subscription to the user's personal channel, so they receive collection updates, according to ACL settings for them.


georgiwe:
I amended the expects of existing tests. But all of the live service tests are being skipped - they have not been "migrated to the mono repo". They use es7 imports, paths have not been changed, etc.

tejasranade:
@thomasconner @georgiwe it looks like a couple of updates have been pushed up to this PR since the last comment. Does it address the concerns with tests? Is more required on this PR before merge?

thomasconner:
Some tests are working for Live Service but not all of them. I would need to spend more time with them but given that @georgiwe created them before he might be able to get them working quicker then I would. This isn't really an issue with the PR so it doesn't appear that there is more required for this PR unless we are able to get unit tests to work for just the changes made in this PR and work on getting the rest of the live service unit tests running in a separate PR.

georgiwe:
While running the tests for live service is not part of this PR, they are part of the mono repo story.

And since I'm not a fan of doing things twice, I'd like to do them once we've settled on infrastructure for running the tests. I've logged [MLIBZ-2245](https://kinvey.atlassian.net/browse/MLIBZ-2245) for "test infrastructure" and I'd rather we did this after that, which I think should also be part of the move to mono repo.

PR #174, opened 2017-12-01
MLIBZ-2197 Spike integration datastores

georgiwe:
#### Description
The first part of the migration of offline redesign POC.

#### Changes
Introduced datastore changes and brought back babel and linting.


georgiwe:
The code in data stores predates the operation factory. The plan was to move to it and use it in each method of data store, and remove the `_buildOperationObject` method.

```javascript
const operation = this._buildOperationObject(OperationType.Read, query);
return this._executeOperation(operation, options);

// becomes
const operation = opFactory.getOperationWithQuery(OperationType.Read, this.collection, query);
return this._executeOperation(operation, options);
```

This would result in using `this.collection` in multiple places. The factory gives us a cleaner interface for creation of Operation objects - not having to pass null in the correct place in its constructor. Using the factory would also facilitate testing - to test if the correct operations are being created, we can mock the operation factory. But if we are mocking the processor, we can check what argument is passed to its `process()` method anyways, so I think both ways are viable. Do you still prefer the instatiation?

Consolidating data stores is a step I had planned for later on, once we implement everything, since it won't change anything below that layer so we don't have to think about it yet. Data stores require a lot of refactoring, but this particular part I wanted to postpone as much as possible.

Exposing callbacks for CacheStore cache results would be a breaking change and is definitely out of scope, but I'd like to see the observables gone too. As you may have noticed they make things difficult in `CacheOfflineProcessor` and break the Liskov principle and if this wasn't JavaScript, I'd be in trouble :D Anyways, if @tejasranade doesn't mind, we can do it as part of the redesign, but I doubt he'll be up for it. And I guess it's not really necessary at this point. I'd rather get it done sooner, so we can move to testing sooner. But the integration tests are not ready yet anyways, so... Who knows :) But we do have quite a bit of work ahead of us on this as it is.

PR #176, opened 2017-12-01
MLIBZ-2197 Add repositories.

georgiwe:
#### Description
Third step - repositories.

#### Changes
Add repositories, fix paths, etc.


packages/kinvey-datastore/src/repositories/network-repository.js
```diff
@@ -0,0 +1,99 @@
...
+
+  count(collection, query, options) {
+    const requestConfig = this._buildRequestConfig(collection, RequestMethod.GET, null, query, '_count', options);
+    return this._makeHttpRequest(requestConfig);
+  }
+
+  _processBatch(collection, method, entities, options) {
+    const isSingle = !Array.isArray(entities);
+    const requestPromises = ensureArray(entities).map((entity) => {
+      const id = method === RequestMethod.PUT ? entity._id : null; // TODO: this isn't great :)
```
tejasranade: Do you mean the logic, the syntax, or both? :)
The logic looks suspect. We support customers setting their own `_id` value on new created entities. This code seems to erase it.

packages/kinvey-datastore/src/repositories/network-repository.js
```diff
@@ -0,0 +1,99 @@
...
+      const requestConfig = this._buildRequestConfig(collection, method, entity, null, id, options);
+      return this._makeHttpRequest(requestConfig);
+    });
+    // TODO: different error handling?
+    return Promise.all(requestPromises)
+      .then(res => (isSingle ? res && res[0] : res));
+  }
+
+  _makeHttpRequest(requestConfig, deltaFetch) {
+    if (deltaFetch) { // TODO: maybe this needs to be in sync manager
```
tejasranade: Needs to be thought out. This is too deep in the stack for the delta set logic. I'd expect it to be implemented in the "processor", since it utilizes both the cache and the network. Also, `_makeHttpRequest` isn't the right place anyway, since even `delete` seems to use it.

packages/kinvey-datastore/src/repositories/network-repository.js
```diff
@@ -0,0 +1,99 @@
...
+   * @param {RequestOptions} options
+   * @param {RequestMethod} method
+   * @param {Objet} data
+   * @returns {RequestOptions}
+   */
+  _buildRequestConfig(collection, method, data, query, idOrRestAction, options) {
+    options = options || {};
+    const config = {
+      method,
+      pathname: buildCollectionUrl(collection, idOrRestAction),
```
tejasranade: `idOrRestAction` sounds very loaded for a method parameter.

packages/kinvey-datastore/src/repositories/offline-repositories/inmemory-offline-repository.js
```diff
@@ -0,0 +1,203 @@
...
+    return this._getAllCollections()
+      .then((collections) => {
+        const promises = collections.map(c => this._deleteAll(c));
+        return Promise.all(promises);
+      })
+      .then(() => this._deleteAll(masterCollectionName));
+  }
+
+  // ----- protected methods
+  _ensureCollection(collection) {
```
tejasranade: This method name doesn't say much about what it is supposed to do. A comment or a better name would help.

packages/kinvey-datastore/src/repositories/offline-repositories/inmemory-offline-repository.js
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  // ----- public methods
+
+  // TODO: add checks for existing collection, try/catch, etc - here or in processor
+
+  create(collection, entitiesToSave) {
+    return this._enqueueCrudOperation(collection, () => {
+      return this._ensureCollection(collection)
+        .then(() => this._readAll(collection))
```
tejasranade: Seems highly inefficient. Good place to target for subsequent performance improvement.

packages/kinvey-datastore/src/repositories/repository-provider.js
```diff
@@ -0,0 +1,47 @@
...
+// TODO: all inmemory instances should share the queue. are there better ways to share it?
+// queue needs to be by collection
+const queue = new InmemoryCrudQueue();
+// TODO: not great to do this here, but tests fail otherwise
+let supportedStorages = {
+  [storageType.inmemory]: inmemoryRepoBuilder,
+  [defaultConstructorName]: inmemoryRepoBuilder
+};
+
+function getRepoType() {
```
tejasranade: Seems like a redundant method. All you are doing is returning the `storageType` from `Client`. The `getOfflineRepository` method below can directly call that.

packages/kinvey-datastore/src/repositories/network-repository.js
```diff
@@ -0,0 +1,99 @@
...
+ * @property {Object} [options.aggregation]
+ * @property {Number} [options.timeout]
+ * @property {Object} [options.properties]
+ * @property {Boolean} [options.skipBL]
+ * @property {Boolean} [options.trace]
+ */
+
+export class NetworkRepository extends Repository {
+  read(collection, query, options) {
+    const requestConfig = this._buildRequestConfig(collection, RequestMethod.GET, null, query, null, options = {});
```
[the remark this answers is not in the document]
georgiwe: The won't be ignored - they are all passed to `_buildRequestConfig()` as its last argument. The separate flag for `_makeHttpRequest()` is the way I reuse deltaset fetch - it's a flag for it to instantiate a `DeltaFetchRequest`, rather than a `KinveyRequest` for the network call. But it supplies all options. Have a look at the `_buildRequestConfig()` method

packages/kinvey-datastore/src/repositories/network-repository.js
```diff
@@ -0,0 +1,99 @@
...
+
+  _processBatch(collection, method, entities, options) {
+    const isSingle = !Array.isArray(entities);
+    const requestPromises = ensureArray(entities).map((entity) => {
+      const id = method === RequestMethod.PUT ? entity._id : null; // TODO: this isn't great :)
+      const requestConfig = this._buildRequestConfig(collection, method, entity, null, id, options);
+      return this._makeHttpRequest(requestConfig);
+    });
+    // TODO: different error handling?
+    return Promise.all(requestPromises)
```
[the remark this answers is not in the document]
georgiwe: There are no restrictions. If someone wants to make 1000 requests, they should be allowed to, imo. It's their problem how much memory their app is allowed to use, or is available, or how the platform manages http connections or a number of variables which are outside our control and if we introduce this sort of throttling, we are limiting the user's options without knowing enough about their setup. And it's not difficult to manage the request count on their own. I think we may introduce such throttling for something that would cause our backend to choke, if there is such a thing, but something that would handle client-side issues.. I think we can let the developer manage that.

Also, I haven't seen such a mechanism in the old design too. There was batching, but it was for push only. It did 100 at a time, I believe. If you insist, we can add it, it's easy.

packages/kinvey-datastore/src/repositories/network-repository.js
georgiwe: I meant that it isn't apparent what is going on by reading it. "if it's a PUT, do it by id". And it has tripped you up, it seems.

It doesn't take away the ability to create your own ids. If you do an update, it will do a PUT to the URL, slash id (which is an upsert for the backend and it works). If you are doing a create, it does a POST with whatever you give it. It shouldn't be removing anything - what do you mean?

packages/kinvey-datastore/src/repositories/network-repository.js
georgiwe: This is temporary - it's how I reuse deltaset fetching. I can refactor it a bit, but we can reuse all the logic for deltaset and it is out of scope for this, so I think this is the best way to go about it. This `deltaFetch` is the flag I was talking about in the first comment, about it being the only option being passed along.

As for the eventual place for it, I thought it's in the `SyncManager` class, as I believed it was only relevant for pull. But it turns out it's also relevant for `CacheStore.find()`. It used to be in `NetworkStore.find()` too, but I think it shouldn't be there. So the minimum change required at this point is to add support for deltaset fetching for `CacheStore.find()`. I think the rest can be left for when we start optimizing for performance, as deltaset needs to be reworked in general. Up to you, really.

packages/kinvey-datastore/src/repositories/network-repository.js
georgiwe: This is either an entity id or something like `_count`. It's not great, I agree :) I guess I should separate those into two. I don't think there are any "by id" actions in the API, but still, I will separate those into two.

packages/kinvey-datastore/src/repositories/offline-repositories/inmemory-offline-repository.js
georgiwe: 👍 
Will rename it to `_ensureCollectionExists()`.

packages/kinvey-datastore/src/repositories/offline-repositories/inmemory-offline-repository.js
georgiwe: Could you elaborate?

packages/kinvey-datastore/src/repositories/repository-provider.js
georgiwe: This is the last thing I was working on before we decided to start moving code. It's the part I mentioned needs to be polished. I wasn't/am not very sure what the best way to achieve this is yet, so I'm separating actions into functions. I don't think leaving it this way is too bad either, though. Other functions not caring where the type comes from, naming the operation, bridging setting name and actual logic. Will improve during development.

packages/kinvey-datastore/src/datastores/datastore.js
```diff
@@ -5,8 +5,8 @@ import { isDefined } from 'kinvey-utils';
 import NetworkStore from './networkstore';
 import CacheStore from './cachestore';
 import SyncStore from './syncstore';
-// import { processorFactory } from '../processors';
-// import { repositoryProvider } from '../repositories';
+import { processorFactory } from '../processors';
+import { repositoryProvider } from '../repositories';
```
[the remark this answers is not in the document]
georgiwe: The `DataStore` class isn't a data store in the sense that `SyncStore`, `CacheStore` and `NetworkStore` are, they do not inherit it. It's merely a factory for them.

Turns out, the `DataStore` class has a static `clear()` method, which deletes all data. The reason I can't instantiate, say, a `SyncStore` and clear its data, is that it clears data only for the collection it is created for and it doesn't know what other collections exist. I can't think of a better way of achieving this.

packages/kinvey-datastore/src/repositories/network-repository.js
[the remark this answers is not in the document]
tejasranade: You are right. I didn't notice that the last parameter has a default empty value.

packages/kinvey-datastore/src/datastores/datastore.js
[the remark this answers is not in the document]
tejasranade: I haven't dug into the details yet, but isn't there a way to delegate this to the processor instead?

packages/kinvey-datastore/src/repositories/network-repository.js
tejasranade: I guess the code is confusing - from just reading it here, it looks like this piece of code will preserve an `id` *only* in the case of `PUT`, otherwise discard it. From your explanation, that seems to not be the case.

I'm ok leaving it as-is. Maybe add a comment describing what's going on.

packages/kinvey-datastore/src/repositories/network-repository.js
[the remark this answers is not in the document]
tejasranade: The point is not to disallow the user from doing something. On the contrary - we are exposing an API which takes an open-ended parameter (a list of items with no limits). We want to make sure the API functions under various cases of input, or fails predictably. Particularly since this is a `save()`, we have prior experience where users tried to save a very large list of items.

I understand that the previous implementation didn't handle this case for the `save()` API, but it should have. The batching on `push()` was added for the exact same reason. Since this is a net new enhancement to the SDK, I am ok with adding a separate ticket to address it - as long as we agree it is necessary.

Effectively, the question worth asking is this  - if I call a `save()` with 1k entities, what is supposed to happen? What about 10k? What is a reasonably large number that we can support by adding logic to batch requests?

packages/kinvey-datastore/src/repositories/network-repository.js
tejasranade: Agree with the second part. Delta set logic shouldn't be in the `NetworkStore`. 

As for your comment about when we refactor it to its "eventual" solution - please open a JIRA ticket with your comment as the context. These are the types of gaps in our current implementation that I'm hoping to uncover during the redesign.

packages/kinvey-datastore/src/repositories/offline-repositories/inmemory-offline-repository.js
tejasranade: The fact that we call `_readAll` at multiple places in this class seems inefficient. I know it's not in scope for the redesign, but it points to a performance issue that'll need to be solved eventually.

packages/kinvey-datastore/src/repositories/repository-provider.js
tejasranade: I didn't fully understand your comment, but I'll wait for the improvements and review again :)

packages/kinvey-datastore/src/datastores/datastore.js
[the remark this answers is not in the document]
georgiwe: Processors also work on a "by collection" basis. The first class in the hierarchy, which knows what collections exist for a particular app key is the repository. We can, of course, do anything. The offline repositories know how to handle a clear when no collection is passed - we can have a separate processor method (this would require a separate operation type), or a setting in the options object of `clear()`, for the processor to know to call `clear()` without the collection parameter (hacky).

Why have it in a processor? Repositories are a data access layer and as such can be used by any non-public class. I don't think it's that bad to have it here. It's not worse than adding an operation type, in order to be able to instantiate a processor (since this is a static method on the `Kinvey.DataStore` object and no data store is instantiated at that point), just to call one of its methods.

packages/kinvey-datastore/src/repositories/network-repository.js
georgiwe: I think this confusion stems from looking at a PR/diff, rather than the actual code.

This code only builds the `KinveyRequest` config object. It doesn't alter the entities being sent to the backend, it just looks for the `_id` property on the entity when doing a put, in order to make sure the correct URL is formed.

packages/kinvey-datastore/src/repositories/network-repository.js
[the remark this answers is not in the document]
georgiwe: Actually, neither `create()`, nor `update()` allow for multiple inserts. They only allow for one entity to be passed.

I'm sorry, if I misunderstood. This is present in the old code and the new. I meant that the internals have no restrictions. Things like the 10k max page size don't apply offline and I've kept all repositories able to do batch operations. Validations for such things are done at the topmost level - in the data stores.

Internally, processors and such may ask repositories to do bulk create and update. Repositories do these in the optimal way for them. `NetworkRepository` does it one by one. It's done as it's told, as best as it can. Whoever is ordering the work being done should mind these considerations.

So, to sum up :)
In the past, restrictions for a per-entity `create` and `update` were in place, and this remains the case. I believe this is the same as what I said in the first post - client is allowed to make 10000 simultaneous requests. But we also expose an API which functions under various input - just the input is very limited :)
The part where we need to be careful is internals - there are no restrictions there. I think the only place where this may be a problem is during `push()` and it's handled there - kept it as it was, 100 entities at a time.

How do you like this setup?

packages/kinvey-datastore/src/repositories/network-repository.js
georgiwe: I meant more the functionality, than the logic code. I mean NetworkStore should not be able to do a deltaset find. And I think we all agree on that.

Where the logic goes would become apparent when we decide on who should be able to use it. And I think only `pull()` should use it.

Should `CacheStore.find()` be able to use deltaset fetching? What should the response of that be - should it only contain the changed entities, or all entities matching the query?

`CacheStore`'s responses are still a bit of a mystery to me. I view it as a `NetworkStore`, which first gives you the results currently in the cache, and then updates the cache with whatever it got from network. Is that a fair assumption? Is this the way it should throw errors - only based on network operations? I guess in this sense, it shouldn't be able to do a deltaset `find`. Perhaps this should be an internal optimization on how it does a `find`, but shouldn't be a public API/setting?

packages/kinvey-datastore/src/repositories/offline-repositories/inmemory-offline-repository.js
georgiwe: This being an "inmemory" repository, there is no way around that. Any performance-intensive work should not be using this type of repository at all. But it is there for backward compatibility. The TTL we discussed in another comment was precisely for this. To avoid multiple reads, the read items are being cached inmemory. This is done in the inmemory repo's persister.

packages/kinvey-datastore/src/repositories/repository-provider.js
georgiwe: Deal :)

PR #184, opened 2017-12-07
MLIBZ-2225: datastore.findById() should throw a NotFoundError if the id argument is not defined

thomasconner:
#### Description
Add validation check to `datastore.findById()` for `id` argument. Throw a `NotFoundError` if the `id` argument is not defined.

#### Tests
Updated unit tests for each datastore type to check if a `NotFoundError` is thrown when the `id` argument is not defined.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2630,6 +2630,15 @@
         "mime-types": "2.1.17"
       }
     },
+    "formatio": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/formatio/-/formatio-1.2.0.tgz",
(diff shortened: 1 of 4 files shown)


tsvetomir-nedyalkov:
It seems that this PR fixes a different case which is not a regression -  when findById is called without an argument - storeToTest.findById(). The described case in Steps To Reproduce of MLIBZ-2225 is when the _id does not exist in the backend - storeToTest.findById('not_existing_id').toPromise().

thomasconner:
I misunderstood the ticket and I fixed the WebSQL adapter that contained the error.

PR #185, opened 2017-12-08
MLIBZ-2234: Export the Kinvey namespace as a module for new bundles created from Mono repo

thomasconner:
#### Description
Export the `Kinvey` namespace as a module from each entry file for Webpack in the packages.

#### Tests
No tests were added. I verified that the bundle exported the `Kinvey` namespace as a module by loading the created bundle in a test app and importing the `Kinvey` module.


src/kinvey.d.ts
```diff
@@ -1,11 +1,532 @@
 // Type definitions for Kinvey JavaScript SDK
 // Definitions by: Kinvey <https://github.com/Kinvey>
 
-import { Observable } from 'rxjs/Observable';
+import {
+  Observable
+} from 'rxjs/Observable';
 
-export as namespace Kinvey;
+export namespace Kinvey {
```
tejasranade: Clearly this will work, but it involves more maintaining since we now have 2 copies of exports in the same file. I take that there's no easier way to keep backwards compatibility. 

A couple of follow-ups that'll help:
- Add a comment above this export, briefly describing why it exists and suggesting that it is deprecated. It helps us as well as SDK users who look at our typescript definition.
- If we are recommending the "new" way of importing a definition, a devcenter PR that accompanies this release (or follows soon after) should change the snippets.

thomasconner:
@georgiwe I have already updated `kinvey.d.ts` to reflect the two ways of importing the SDK. I verified this in a NativeScript TypeScript app that I could import it both ways as well.

Our SDK exports a bunch of smaller modules. Since we are deprecating, `import { Kinvey } from 'kinvey-nativescript-sdk'` we will be removing the file that makes this import style possible in the future. To support `import Kinvey from 'kinvey-nativescript-sdk'` we would still need another export file that captures all the modules in an object and exports that object as the default export. `import * as Kinvey from 'kinvey-nativescript-sdk'` provides the same behavior without us having to maintain this second export file. I would also argue that other libraries are doing it wrong and if your library exports multiple modules you should be doing it this way.

PR #186, opened 2017-12-11
Datastores redesign part 1 - preparation

georgiwe:
#### Description
First part of MLIBZ-2238.

#### Changes
Set up linting and fix the more important errors it showed. Add necessary utility functions in relevant places.


.eslintrc
```diff
@@ -12,6 +15,7 @@
     "import/newline-after-import": "off",
     "import/no-extraneous-dependencies": "off",
     "import/prefer-default-export": "off",
+    "linebreak-style": "off",
```
thomasconner: Should we adopt the recommendation outlined [here](https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems) instead of turning the rule off completely?

.eslintrc
```diff
@@ -1,6 +1,9 @@
  {
   "extends": "airbnb-base",
   "parser": "babel-eslint",
+  "env": {
```
thomasconner: It should be possible to extend the .eslintrc file and add this environment for just the unit tests

package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Kinvey, Inc.",
   "scripts": {
     "build": "lerna run build --parallel",
-    "test": "mocha --require babel-polyfill --require babel-register 'src/**/*.spec.js'"
+    "test": "mocha --require babel-polyfill --require babel-register src/**/*.spec.js"
```
thomasconner: This only runs about half the unit tests on MacOS. The solution is use `\"src/**/*.spec.js\"` instead.

src/core/client.js
```diff
@@ -211,8 +212,8 @@ export class Client {
     }
 
     if (timeout < 0) {
-      Log.info(`Default timeout is less than 0. Setting default timeout to ${defaultTimeout}ms.`);
-      timeout = defaultTimeout;
+      Log.info(`Default timeout is less than 0. Setting default timeout to ${this.defaultTimeout}ms.`);
```
thomasconner: Should this use `DEFAULT_TIMEOUT`?

.eslintrc
georgiwe: Hm, why though? I'd rather not mess with git's way of doing it - I like it. "check out with Windows line endings, check in with unix" - I don't have to think about it at all. So I don't need the linting - git handles it. Do you need the linting?

.eslintrc
georgiwe: Yeah, let's do this in master/mono-repo branch even - not wait for the redesign to be merged.

package.json
georgiwe: I'm still not comfortable deciding on this, since neither of us knows what is happening :D

src/core/client.js
georgiwe: I don't think so. The instance value is set in the constructor and that features `DEFAULT_TIMEOUT` as the default or something.

.eslintrc
thomasconner: No I don't think so. I just wanted to ask the question and see if it was considered. I think turning it off will be fine.

package.json
thomasconner: Can we make the change so unit tests continue to work on Windows and MacOS environments? Then we can spend time learning the why.

georgiwe: Sure, but this isn't the place for it. This stuff isn't going to get merged into master for months. I investigated it a bit. I think it comes down to the library mocha uses to expand glob patterns. In the mocha docs it's with double quotes, so I think we're good on that front. We should make the change in the important branches - master, I guess.

src/core/client.js
thomasconner: This is correct but I was thinking about a different use case. For example, I start out by setting the `defaultTimeout` for the `client` instance to `-1` which results in the value being set to `DEFAULT_TIMEOUT`.

```js
const client = new Client({ defaultTimeout: -1 });
// client.defaultTimeout === DEFAULT_TIMEOUT (60000)
```

But later on lets say I change the `defaultTimeout` to 30s.

```js
client.defaultTimeout = 30000
```

And then somewhere else I set the default timeout to -1

```js
client.defaultTimeout = -1;
```

This will set the `defaultTimeout` to the previous value of 30s. Is this the intended result we want or should it set it to `DEFAULT_TIMEOUT`? I think it makes sense to use the previous value and the way it works now is the intended result but I wanted to double check.

I am aware I wrote this code but now I am getting the chance to review it again 😄.

georgiwe: Um... if you pass an invalid parameter, this should result in a runtime exception. I'm not a fan of this sort of fault tolerance :) This is my opinion. But this is out of scope anyways. It can be examined separately from the redesign.

This is just fixing an error which would have resulted in a `ReferenceError`. It was pointed out to me by the linter.

package.json
thomasconner: But if I or someone else was to checkout this branch, work on the code, add unit tests, etc. and then try to run unit tests it won't work. That is why we need to change it here.

src/core/client.js
thomasconner: Ok I will add a JIRA ticket to capture this somewhere else since this is out of scope.

package.json
georgiwe: Yes, we need to do it in both places, because each is broken for one of us :D

.eslintrc
georgiwe: Created a ticket for it - [MLIBZ-2252](https://kinvey.atlassian.net/browse/MLIBZ-2252).

src/core/client.js
georgiwe: I believe [MLIBZ-2251](https://kinvey.atlassian.net/browse/MLIBZ-2251) is the ticket for this.

georgiwe:
The `test` script is changed in a later PR. Merging this one.

PR #187, opened 2017-12-11
Datastores redesign part 2 - persisters

georgiwe:
#### Description
Second part of MLIBZ-2238.

#### Changes
Add persisters.

### DIFF
--- a/src/core/datastore/persisters/key-value-persister.js
+++ b/src/core/datastore/persisters/key-value-persister.js
@@ -1,0 +1,67 @@
+const _cache = {};
+
+export class KeyValuePersister {
+  /** @type {Boolean} */
+  _cacheEnabled;
+  /** @type {Number} */
+  _ttl;
+
+  // TODO: implement TTL, make _cache a constructor argument?
+  constructor(cacheEnabled = true, ttl = Infinity) {
+    this._cacheEnabled = cacheEnabled;
+    this._ttl = ttl;
+  }
+
+  readEntities(key) {
+    if (this._cacheEnabled && _cache[key]) {
+      return Promise.resolve(_cache[key]);
+    }
+
+    return this._readFromPersistance(key)
+      .then((entities) => {
+        entities = entities || [];
+        if (this._cacheEnabled) {
+          _cache[key] = entities;
+        }
+        return entities;
+      });
+  }
+
+  persistEntities(key, entities) {
+    if (this._cacheEnabled) {
+      delete _cache[key];
+    }
+    return this._writeToPersistance(key, entities)
+      .then((result) => {
+        if (this._cacheEnabled && this._ttl < Infinity) {
+          setTimeout(() => {
+            delete _cache[key];
+          }, this._ttl);
+        }
+        return result;
+      });
+  }
+
+  deleteEntities(key) {
+    if (this._cacheEnabled) {
+      delete _cache[key];
+    }
+    return this._deletePersistance(key);
+  }
+
+  _throwNotImplementedError() {
+    throw new Error('Abstract method not implemented');
+  }
+
+  _readFromPersistance(key) {
+    this._throwNotImplementedError(key);
+  }
+
+  _writeToPersistance(key, array) {
+    this._throwNotImplementedError(key, array);
+  }
+
+  _deletePersistance(key) {
+    this._throwNotImplementedError(key);
+  }
+}
--- a/src/core/datastore/utils/utils.js
+++ b/src/core/datastore/utils/utils.js
@@ -1,0 +1,38 @@
+import { Metadata } from '../../metadata';
+import { Client } from '../../client';
+
+export function buildCollectionUrl(collectionName, id, restAction) {
+  let result = `appdata/${Client.sharedInstance().appKey}/${collectionName}`;
+  if (id) {
+    result += `/${id}`;
+  }
+  if (restAction) {
+    result += `/${restAction}`;
+  }
+  return result;
+}
+
+export function generateEntityId(length = 24) {
+  const chars = 'abcdef0123456789';
+  let objectId = '';
+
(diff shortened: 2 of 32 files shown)


src/core/datastore/persisters/key-value-persister.js
```diff
@@ -0,0 +1,67 @@
+const _cache = {};
+
+export class KeyValuePersister {
+  /** @type {Boolean} */
+  _cacheEnabled;
+  /** @type {Number} */
+  _ttl;
+
+  // TODO: implement TTL, make _cache a constructor argument?
```
thomasconner: Let's capture this TODO and remove `ttl` since it isn't being used.

src/core/datastore/persisters/key-value-persister.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  readEntities(key) {
+    if (this._cacheEnabled && _cache[key]) {
+      return Promise.resolve(_cache[key]);
+    }
+
+    return this._readFromPersistance(key)
+      .then((entities) => {
+        entities = entities || [];
```
thomasconner: Instead of this you can do `((entities = []) => {`. This is stylistic but more concise.

src/core/datastore/persisters/key-value-persister.js
```diff
@@ -0,0 +1,67 @@
...
+      });
+  }
+
+  persistEntities(key, entities) {
+    if (this._cacheEnabled) {
+      delete _cache[key];
+    }
+    return this._writeToPersistance(key, entities)
+      .then((result) => {
+        if (this._cacheEnabled && this._ttl < Infinity) {
```
thomasconner: Is `ttl` implemented? I thought from the TODO above that it was not.

src/core/datastore/utils/utils.js
```diff
@@ -0,0 +1,38 @@
+import { Metadata } from '../../metadata';
+import { Client } from '../../client';
+
+export function buildCollectionUrl(collectionName, id, restAction) {
+  let result = `appdata/${Client.sharedInstance().appKey}/${collectionName}`;
+  if (id) {
+    result += `/${id}`;
```
thomasconner: We should use `url-join` which is already a dependency of the repo.

src/core/datastore/utils/utils.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  return objectId;
+}
+
+export function isNotEmpty(object) {
+  return !!object && (object.length > 0 || Object.keys(object).length > 0);
+}
+
+export function isEmpty(object) {
```
thomasconner: Why not use https://lodash.com/docs/#isEmpty?

src/core/datastore/persisters/key-value-persister.js
georgiwe: Well, the class supports it, but nobody is using it. There is a general decision to be made about it, so the todo is there to make sure I don't forget it. There is a brief discussion about it in the previous "set" of PRs. It's still outstanding as to whether to keep it and in what form.

src/core/datastore/persisters/key-value-persister.js
georgiwe: True. It wouldn't overwrite `null`, though. But I had plans to make persisters more "general-purpose" so they shouldn't overwrite `null` anyways. I'll make this change 👍

src/core/datastore/utils/utils.js
georgiwe: Okay, sure 👍

src/core/datastore/utils/utils.js
georgiwe: Hah, yeah... `isNotEmpty()` (above) used to be a different check, but I guess it has grown into lodash's `!isEmpty()`.

👍 

Btw, didn't you say we were moving away from lodash?

thomasconner: I have been looking into ways we can decrease the bundles size of our SDKs. Lodash continues to add a lot of code that we may not need (even with tree shaking). It is quite possible we could write the necessary utility functions as part of the SDK and remove the dependency on Lodash.

The downside to this is Lodash is well tested and we know we can trust it where as rewriting these utilities would require a lot more testing to cover edge cases. 

For now we still have it as a dependency so lets continue to use it until we decide not too.

src/core/datastore/persisters/key-value-persister.js
thomasconner: https://kinvey.atlassian.net/browse/MLIBZ-2236 is the related ticket in JIRA.

georgiwe: The TTL is implemented, just not being used. Since the caching is toggled by a flag, I'll set its default value to false, so it will have no effect. Once we decide something, we can change that, but I think it's best to leave the code there, since it has a switch and will be set to "disabled".

src/core/datastore/persisters/key-value-persister.js
georgiwe: My comment from above, about this TTL, might have been more suitable for over here :)

src/core/datastore/utils/utils.js
georgiwe: This is done in a later PR.

src/core/datastore/utils/utils.js
georgiwe: Done in a later PR.

src/core/datastore/persisters/key-value-persister.js
georgiwe: I removed the default value as it's not the persister's job to decide what the default returned value should be. Whoever is using it should know better. There already is a default value in the inmemory repository (which is the only one using a KeyValuePersister for now) so no other change is necessary, I believe.

PR #197, opened 2017-12-12
MLIBZ-2247:  Add rxjs as a dependency to use in definitions file

thomasconner:


tejasranade:
@thomasconner how was this tested? Do you know if/how this affects existing SDK users?

thomasconner:
@tejasranade I tested it locally with a test app. No this won't affect existing SDK users. At least not in anyway I can think of. Did you have something in mind that concerned you or was it just a question to see if I had thought about it?

PR #201, opened 2018-01-08
MLIBZ-2157: Add filter for a not equal query properly

thomasconner:
#### Changes
I added a condition to the `if` statement in the `addFilter()` function for a query to handle the special use case for `$ne`.

#### Tests
I added a unit tests that verifies that a `$ne` query filters data correctly.

### DIFF
--- a/src/core/query.js
+++ b/src/core/query.js
@@ -665,7 +665,7 @@ export class Query {
    */
   addFilter(field, condition, values) {
     if (isDefined(condition)
-        && (isDefined(values) || condition === '$ne')) {
+      && (isDefined(values) || arguments.length === 3)) {
       if (!isPlainObject(this.filter[field])) {
         this.filter[field] = {};
       }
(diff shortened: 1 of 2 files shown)


src/core/query.js
```diff
@@ -664,7 +664,8 @@ export class Query {
    * @returns {Query} The query.
    */
   addFilter(field, condition, values) {
-    if (isDefined(condition) && isDefined(values)) {
+    if (isDefined(condition)
+        && (isDefined(values) || condition === '$ne')) {
```
georgiwe: It seems to me that each separate filter method should validate the values for itself. If the check is in here, this method needs to know what values are valid for which operator. If we need to allow this for another method, we'd need another "or" statement. Just something to think about, I won't request changes - it's up to you.

thomasconner: Yes this is something I thought about. For now this had the least impact to fix the bug. I will look over it again to see if there is a better way but at first glance I did not see another option.

thomasconner:
I made a change that doesn't look at the condition argument. Please review again. All tests still pass.

georgiwe:
This does seem more... concise, but I think it changes behaviour a bit - now `null` can be set as a value for any operator. If that's intended, then that's fine, I think.

PR #203, opened 2018-01-12
Datastores redesign - Add WebSQL persistance

georgiwe:
#### Description
[MLIBZ-2263](https://kinvey.atlassian.net/browse/MLIBZ-2263) - Added WebSQL persistance. First implemented part of stage 3 of the redesign. For easier review, I'm creating separate PRs for each ticket, into a branch for all of stage 3.

#### Changes
Implement branch of repository/persister combinations for work with key/value stores like IndexedDb. It currently fits the way we use WebSQL and SQLite.General bug fixes. Small improvements on tests.

### DIFF
--- a/src/core/datastore/cachestore.js
+++ b/src/core/datastore/cachestore.js
@@ -154,6 +154,9 @@ export class CacheStore extends NetworkStore {
    * @return  {Promise}                                                         Promise
    */
   pendingSyncCount(query, options) {
+    if (query) {
+      return this.syncManager.getSyncItemCountByEntityQuery(this.collection, query);
+    }
     return this.syncManager.getSyncItemCount(this.collection);
   }
 
@@ -189,7 +192,8 @@ export class CacheStore extends NetworkStore {
    */
   pull(query, options = {}) {
     options = assign({ useDeltaFetch: this.useDeltaFetch }, options);
-    return this.syncManager.getSyncItemCount(this.collection)
+    // TODO: the query issue must be resolved - entity or sync entity
+    return this.syncManager.getSyncItemCountByEntityQuery(this.collection, query)
       .then((count) => {
         if (count > 0) {
           // TODO: I think this should happen, but keeping current behaviour
@@ -234,7 +238,6 @@ export class CacheStore extends NetworkStore {
   }
 
   clearSync(query, options) {
-    // return this.syncManager.clearSync(query, options);
-    return this.syncManager.clearSync(this.collection);
+    return this.syncManager.clearSync(this.collection, query);
   }
 }
--- a/src/core/datastore/persisters/key-value-store-persister.js
+++ b/src/core/datastore/persisters/key-value-store-persister.js
@@ -1,9 +1,13 @@
+import { Promise } from 'es6-promise';
+
 import { Client } from '../../client';
+import { NotFoundError, KinveyError } from '../../errors';
 
 import { KeyValuePersister } from './key-value-persister';
 
 export class KeyValueStorePersister extends KeyValuePersister {
-  get _databaseName() {
+  // protected property
+  get _storeName() {
     return Client.sharedInstance().appKey;
   }
 
@@ -13,17 +17,24 @@ export class KeyValueStorePersister extends KeyValuePersister {
 
   writeEntities(collection, entities) {
     return this._writeEntitiesToPersistance(collection, entities)
-      .then((result) => {
+      .then(() => {
         this._invalidateCache(collection);
-        return result;
+        return true;
       });
   }
 
-  deleteEntities(collection, entityIds) {
-    return this._deleteEntitiesFromPersistance(collection, entityIds)
-      .then((result) => {
+  deleteEntity(collection, entityId) {
+    return this._deleteEntityFromPersistance(collection, entityId)
+      .then((deletedCount) => {
         this._invalidateCache(collection);
-        return result;
+
+        if (deletedCount === 0) {
+          return Promise.reject(new NotFoundError(`Entity with id ${entityId} was not found`));
+        }
+        if (deletedCount > 1) {
+          return Promise.reject(new KinveyError('Delete by id matched more than one entity'));
+        }
+        return true;
       });
   }
 
@@ -32,7 +43,7 @@ export class KeyValueStorePersister extends KeyValuePersister {
     this._throwNotImplementedError(entities);
   }
 
-  _deleteEntitiesFromPersistance(collection, entityIds) {
+  _deleteEntityFromPersistance(collection, entityIds) {
     this._throwNotImplementedError(entityIds);
   }
 }
--- a/src/core/datastore/persisters/key-value-store-persister.js
+++ b/src/core/datastore/persisters/key-value-store-persister.js
@@ -7,15 +7,32 @@ export class KeyValueStorePersister extends KeyValuePersister {
     return Client.sharedInstance().appKey;
   }
 
-  readEntity(key, entityId) {
-    this._throwNotImplementedError();
+  readEntity(collection, entityId) {
+    this._throwNotImplementedError(entityId);
   }
 
-  writeEntity(key, entity) {
-    this._throwNotImplementedError();
+  writeEntity(collection, entity) {
+    return this._writeEntityToPersistance(collection, entity)
+      .then((result) => {
+        this._invalidateCache(collection);
+        return result;
+      });
   }
 
-  deleteEntity(key, entityId) {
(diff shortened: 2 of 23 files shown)


src/core/datastore/persisters/key-value-store-persister.js
```diff
@@ -0,0 +1,49 @@
+import { Promise } from 'es6-promise';
+
+import { Client } from '../../client';
+import { NotFoundError, KinveyError } from '../../errors';
+
+import { KeyValuePersister } from './key-value-persister';
+
+export class KeyValueStorePersister extends KeyValuePersister {
```
tejasranade: I'm confused by this class. What is it supposed to represent? What does it add to the `KeyValuePersister`? Other than the `_storeName`, I don't see any particular responsibility for this class.

src/core/datastore/processors/cache-offline-data-processor.js
```diff
@@ -62,6 +63,7 @@ export class CacheOfflineDataProcessor extends OfflineDataProcessor {
 
   _processCreate(collection, data, options) {
     let offlineEntity;
+    data = clone(data);
```
tejasranade: Looks odd. Are you sure you meant this?

src/core/datastore/processors/cache-offline-data-processor.js
```diff
@@ -111,16 +113,12 @@ export class CacheOfflineDataProcessor extends OfflineDataProcessor {
   }
 
   _processUpdate(collection, data, options) {
-    let offlineEntity;
     return super._processUpdate(collection, data, options)
-      .then((result) => {
-        offlineEntity = result;
-        return this._networkRepository.update(collection, data, options);
-      })
+      .then(() => this._networkRepository.update(collection, data, options))
```
tejasranade: Is this updating on the network first, followed by locally? If network fails, is the change persisted locally? (it should be) 
How does it compare to current behavior?

src/core/datastore/processors/offline-data-processor.js
```diff
@@ -63,7 +64,9 @@ export class OfflineDataProcessor extends DataProcessor {
   _processClear(collection, query, options) {
     return this._syncManager.clearSync(collection, query)
       .then(() => this._getRepository())
-      .then(repo => repo.clear(collection, query, options));
+      // TODO: from the public API, it seems that clear is just a delete with no filter
```
tejasranade: Re: your comment - the difference between `clear` and a `delete` is the following:
- A `delete` also implies that the entities will be deleted from the backend (in case of a sync store, this implies we'll keep track of the `delete` in the sync queue and apply it to the backend on a subsequent `push` or `sync`).
- A `clear` is only supposed to clear local data. It has no effect on the backend. A `clear` does not impact the sync queue.

src/core/datastore/processors/offline-data-processor.js
```diff
@@ -105,10 +108,14 @@ export class OfflineDataProcessor extends DataProcessor {
   }
 
   _addMetadataToEntities(data) {
-    ensureArray(data).forEach((entity) => {
+    const siSingle = !Array.isArray(data);
```
tejasranade: Assuming this is checking whether data is a single entity or an array - wouldn't you want to add the metadata in case it is a single entity?

tejasranade: On a second look, maybe your `ensureArray()` converts a single entity into an array of 1.
A little confusing, nevertheless.

src/core/datastore/cachestore.js
```diff
@@ -154,6 +154,9 @@ export class CacheStore extends NetworkStore {
    * @return  {Promise}                                                         Promise
    */
   pendingSyncCount(query, options) {
+    if (query) {
+      return this.syncManager.getSyncItemCountByEntityQuery(this.collection, query);
```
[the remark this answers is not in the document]
georgiwe: Some time ago, I raised the question of querying the sync queue and the way it works. Because essentially when you are using methods like `pendingSyncCount`, `clearSync` and other sync queue related methods, they all accept a query. And that query doesn't regard the actual sync items, but the entities they are about. So it is applied to the collection (of entities) and whatever entities match, a query to the sync items collection (aka the sync queue) is made, for items which are about those entity ids. There is a problem with this, that deleted entities would not get matched, as they are not in their collection. So the sync items for those entities will not get "used". That's a bug. We don't yet have a solution for this. In order to not change functionality, I've kept those public methods working as they used to, but introduced this distinction. Where it says "entity query", it's a query which works on entities, because I think if you just call it query, one might assume it works on sync items.

The method `getSyncItemCount` doesn't take a query, because I didn't want to mix the functionality which should get fixed, with the functionality that works as expected - it just gets all the sync items (including ones for deleted entities).

src/core/datastore/processors/cache-offline-data-processor.js
georgiwe: This is to avoid mutating the objects that are passed by the user. What do you find odd?

src/core/datastore/processors/cache-offline-data-processor.js
georgiwe: It updates offline first, then over the network. The offline update is right above - the `super._processUpdate()` part. That's the `_processUpdate()` of `OfflineDataProcessor`. I believe it's the same as the current behaviour.

src/core/datastore/processors/offline-data-processor.js
georgiwe: That is what `ensureArray()` does, yes. Which part is confusing? You ensure something is an array, then you call a `map()` on that array, deep clone each entity and add metadata to it. In the end you return the result in the expected form (whatever was passed), one entity or an array of entities.

src/core/datastore/persisters/key-value-store-persister.js
georgiwe: This is a persister, which can persist to key-value stores. IndexedDb is a key-value store. We are also using WebSQL and SQLite as key-value stores. What I mean by that is that we are writing and reading by key, the id of the entitiy, and the value being the entire entity (in SQL the entity is stringified).

These stores handle "byId" operations efficiently. However, we have querying which has to be done in memory (for `find()`, `remove()`, `count()`). So we need an API, which can, as efficiently as possible, read/write entire collections, as well as single entities from these collections.

The `KeyValuePersister` can read/write/delete by key and we use that for the "entire collection" part. `KeyValueStorePersister` extends on that, by adding methods for doing the same, on an entity level, thus utilizing the capabilities of a key/value store. It encapsulates working with the store entirely, abstracting away the database name (`_storeName` you mentioned) and working with the collections/entities in the most efficient way possible (all in a single transaction, etc).

Ontop of `KeyValueStorePersister` is the `KeyValueStoreOfflineRepository`, which extends upon `InmemoryOfflineRepository` and adds better performance for "byId" operations, utilizing the `KeyValueStorePersister`. It keeps the functionality from its parent, when it comes to querying operations, but overrides "byId" operations. Reusing the inmemory CRUD queue from parent, etc.

src/core/datastore/processors/offline-data-processor.js
georgiwe: I meant it more from an internal standpoint. Because `offlineRepo.clear()` can be more performant than an `offlineRepo.delete()`, but in order to return the count, currently I do a `delete()`.

The second part may be a misunderstanding. Currently, I clear the sync queue. Why leave the sync queue, if we are deleting the entities? We can't possibly sync the updates or creates (also, for create the sync queue is being cleared in the current implementation) upon push, so what is the use of the sync items?

src/core/datastore/cachestore.js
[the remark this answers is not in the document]
tejasranade: That distinction makes sense. "Pushing a subset of changes" is something we've been asked about, which is why the methods with queries exist. 
I see the problem with deletes. As you said, separate topic :)

src/core/datastore/processors/offline-data-processor.js
tejasranade: I think the part that always confuses me is the use of the word `ensure`. Does `ensureArray(data)` imply we're validating that `data` is an array, or does it mean we'll convert `data` to an array if it's not an array? 😕

georgiwe: We're making sure it's an array :) Meaning we'll convert it. Looking at the code would clarify in a second, but reading it as text does allow for some ambiguity. `putInArrayIfNotAnArray()`? :) I'm open to suggestions.

PR #205, opened 2018-01-16
MLIBZ-2156: Correctly sort fields that have a null value

thomasconner:
#### Description
Put docs with the sort field set to `null` at the beginning of the sorted array.

#### Tests
- Added test that checks that fields with a `null` value are sorted to the beginning of the list.

### DIFF
--- a/src/core/query.js
+++ b/src/core/query.js
@@ -783,9 +783,9 @@ export class Query {
             } else if (isDefined(aField) === false && isDefined(bField)) {
               return -1 * modifier;
             } else if (typeof aField === 'undefined' && bField === null) {
-              return -1 * modifier;
+              return 0;
             } else if (aField === null && typeof bField === 'undefined') {
-              return 1 * modified;
+              return 0;
             } else if (aField !== bField) {
               return (aField < bField ? -1 : 1) * modifier;
             }
--- a/src/core/query.js
+++ b/src/core/query.js
@@ -778,9 +778,9 @@ export class Query {
             const bField = nested(b, field);
 
             if (isDefined(aField) && isDefined(bField) === false) {
-              return -1;
-            } else if (isDefined(bField) && isDefined(aField) === false) {
               return 1;
+            } else if (isDefined(bField) && isDefined(aField) === false) {
+              return -1;
             } else if (aField !== bField) {
               const modifier = json.sort[field]; // 1 or -1.
               return (aField < bField ? -1 : 1) * modifier;
--- a/src/core/query.js
+++ b/src/core/query.js
@@ -776,13 +776,17 @@ export class Query {
             // Find field in objects.
             const aField = nested(a, field);
             const bField = nested(b, field);
+            const modifier = json.sort[field]; // 1 (ascending) or -1 (descending).
 
             if (isDefined(aField) && isDefined(bField) === false) {
-              return 1;
-            } else if (isDefined(bField) && isDefined(aField) === false) {
-              return -1;
+              return 1 * modifier;
+            } else if (isDefined(aField) === false && isDefined(bField)) {
+              return -1 * modifier;
+            } else if (typeof aField === 'undefined' && bField === null) {
+              return -1 * modifier;
+            } else if (aField === null && typeof bField === 'undefined') {
+              return 1 * modified;
             } else if (aField !== bField) {
-              const modifier = json.sort[field]; // 1 or -1.
               return (aField < bField ? -1 : 1) * modifier;
             }
           }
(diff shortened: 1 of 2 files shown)


src/core/query.js
```diff
@@ -778,9 +778,9 @@ export class Query {
             const bField = nested(b, field);
 
             if (isDefined(aField) && isDefined(bField) === false) {
```
georgiwe: Wouldn't this treat `null` and `undefined` the same? I'm not sure what the spec is - what should happen to objects, which do not posses the field? Regardless of that, wouldn't this result in an unpredictable order among entities for which `isDefined()` returns false?

src/core/query.spec.js
```diff
@@ -934,6 +934,20 @@ describe('Query', () => {
...
     });
+
+    it('should put docs with null values for sort field at the beginning of the list', () => {
+      const entity1 = { _id: 1, customProperty: randomString() };
+      const entity2 = { _id: null, customProperty: randomString() };
+      const entity3 = { _id: 2, customProperty: randomString() };
+      const entity4 = { _id: 3, customProperty: randomString() };
+      const query = new Query().ascending('_id');
+      query.fields = ['customProperty'];
+      const result = query.process([entity4, entity1, entity3, entity2]);
```
georgiwe: It would help to define the spec more thoroughly to add an entity which does not contain a `someProperty` field at all. Or maybe an object, which is a primitive even - are those treated the same as if they are objects, which don't have it?

src/core/query.spec.js
```diff
@@ -934,6 +934,20 @@ describe('Query', () => {
       expect(result[0].customProperty).to.equal(entity1.customProperty);
       expect(result[1].customProperty).to.equal(entity2.customProperty);
     });
+
+    it('should put docs with null values for sort field at the beginning of the list', () => {
```
georgiwe: Maybe add a test for `descending()` too?

src/core/query.js
thomasconner: Yes this does treat `null` and `undefined` the same (or appears too). When I tested it with `null` and `undefined`, `null` came first in an ascending order and `undefined` came first in a descending order.

The same order would be returned depending on if the sort is `descending` or `ascending` and `isDefined()` returned `false`.

georgiwe: Isn't that because of the order in the initial array? Because comparing `null` and `undefined` always returns `false`. And what is the desired behaviour here? We are trying to align it to the server's, right - what is that?

thomasconner: I will have to see what the server returns for these different use cases. I don't think there is a spec somewhere to reference.

thomasconner: For a sort on `_id` in descending order the server returns

```json
[ 
  { _id: 2, customProperty: 'Xlv3a36WdCahHscBdu' },
  { _id: 1, customProperty: 'UbiUgNfxLdiHrX8vOS' },
  { _id: null, customProperty: 'ul29WKATNXhGD9SUwF' }, // _id is null
  { customProperty: 'yRNvuXJ79vC3OTW9HZ' } // _id is undefined
]
```

For a sort on `_id` in ascending order the server returns

```json
[ 
  { customProperty: 'uhljZH0VRHUNbW9Ck2' }, // _id is undefined
  { _id: null, customProperty: 'YCtYEPDQYyPtnN8n3E' }, // _id is null
  { _id: 1, customProperty: 'oYYRouQ9eZ026J3pKq' },
  { _id: 2, customProperty: 'gUpApX6uCIjYg8FTx8' } 
]
```

thomasconner:
#### Changes
- Changes sorting to return the same result as the server for `null` and `undefined` values

src/core/query.spec.js
```diff
@@ -934,6 +934,21 @@ describe('Query', () => {
...
+
+    it('should put docs with null or undefined values for sort field at the beginning of the list', () => {
+      const entity1 = { _id: 1, customProperty: randomString() };
+      const entity2 = { _id: null, customProperty: randomString() };
+      const entity3 = { _id: 2, customProperty: randomString() };
+      const entity4 = { customProperty: randomString() };
+      const query = new Query().ascending('_id');
+      // query.fields = ['customProperty'];
+      const result = query.process([entity4, entity1, entity3, entity2]);
+      console.log(result);
```
georgiwe: This should probably be removed. And what about the commented line above?

src/core/query.js
georgiwe: How did you create an entity without an _id? I tested it with a different field and null always came up before objects which do not posses the field in ascending order - in both mongodb and for requests to Kinvey.

thomasconner:
#### Tests
- Remove `console.log` statement and uncommented the line

thomasconner:

thomasconner:
@tejasranade Can you review this and approve it if everything looks good?

src/core/query.js
georgiwe: Ah, nevermind, turns out it's just the natural order that's at play here.

thomasconner: The natural order is preserved with the latest changes.

PR #211, opened 2018-01-19
Datastores redesign - Add IndexedDb persistance

georgiwe:
#### Description
MLIBZ-2262 Add necessary classes and logic, in order to work with data in IndexedDb (browser and cordova).

#### Changes
Implement IndexedDbKeyValueStorePersister to work with data in IndexedDb. Small changes in WebSQL and KeyValueStore persisters. Tiny fix for a test to avoid it timing out upon fail.


thomasconner:
The code looks fine. My concern is that we have added this persister to the core but it is only usable in an browser environment which will leave to unnecessary code in bundles for environments that will never be used.

I am approving but just wanted to make that comment.

PR #212, opened 2018-01-19
Datastores redesign - Add LocalStorage and SessionStorage persistance

georgiwe:
#### Description
MLIBZ-2264 Add ability to work with data in LocalStorage and SessionStorage (browser and cordova).

#### Changes
Add necessary classes and remove outdated classes/files.


thomasconner:
The code looks good. My concern is that we have added these persisters to the core but are only usable in an browser environment which will leave to unnecessary code in bundles for environments that will never be used.

I am approving but just wanted to make that comment.

PR #213, opened 2018-01-22
MLIBZ-2120: Add support for specifying a storage adapter

thomasconner:
#### Changes
Added a `storageAdapters` option to `init()` that can be used to select which storage adapter should be used.

```javascript
Kinvey.init({
  appKey: 'appKey',
  appSecret: 'appSecret',
  storageAdapters: Kinvey.StorageAdapter.WebSQL // This can be an array of storage adapters or a single value
});
```

#### Tests
This was tested manually. We should have @tsvetomir-nedyalkov tests this with integration tests.


georgiwe:
I discussed this some with Lyubo and we think it may be worth holding back on the priority list functionality for now. We're not sure it's all that useful and nobody seems to have asked for it yet. And if we release it, we have to support it. How about we just release the single provider selection option for now, and we can think about the list a bit longer?

Here is an example configuration object. Most of it is for the future, but it's the motivation behind having it as an object at an early stage.

```javascript
const StorageProvider = Kinvey.DataStore.StorageProvider;
const SchemaConstants = Kinvey.DataStore.SchemaConstants;

const storageConfig = {
  provider: StorageProvider.WebSql, // this is all we are currently implementing
  providerConfig: {
    [StorageProvider.WebSql]: {
      maxSize: 1024 * 1024,
      indexes: {
        Books: {
          Title: true,
          ISBN: SchemaConstants.Index.Unique
        }
      }
      // ...
    }
  },
  collectionsSchema: { // we could check out how Mongoose.js does it
    Books: {
      Title: SchemaConstants.Text,
      Author: {
        type: SchemaConstants.Relation,
        targetCollection: 'Authors'
      },
      ISBN: SchemaConstants.Text
    }
  }
};

Kinvey.init({
  appKey: 'some string',
  appSecret: 'another string',
  storage: storageConfig
});
```

thomasconner:
1) I think `Kinvey.StorageAdapters` works well but I will change it to something you mentioned.

2) Same thing as above.

3) The enum `Kinvey.StorageAdapters` only contains valid values for the environment you are using. We can validate they don't use a random value that is not in the enum.

4) I disagree with this. If you put it under `Kinvey.DataStore` then it implies you are setting the storage adapter only for `Kinvey.DataStore`. Right now this is the only place we use storage but in the future we might use storage in other areas of the SDK.

thomasconner:
What prevents us from supporting a priority list functionality now? Have we discussed the configuration object and I missed it?

thomasconner:
- I changed `Kinvey.StorageAdapter` to `Kinvey.StorageProvider`.
- I changed the config option for `init()` to `config.storage`.
```javascript
Kinvey.init({
  appKey: 'appKey',
  appSecret: 'appSecret',
  storage: Kinvey.StorageProvider.WebSQL // or [Kinvey.StorageProvider.WebSQL, Kinvey.StorageProvider.IndexedDB]
});
```

thomasconner:
@georgiwe Can you look over the changes I made?

thomasconner:

thomasconner:
@tejasranade Look at the example code snippet and let me know your thoughts. Should we rename the config options `storage` to something else? Since we refer to this as cache elsewhere should we refer to it as `cachePersistence` here?

Should the enum `StorageProvider` be under the `Kinvey.DataStore` or `Kinvey`? Should it be renamed to something with cache in the name as described above?

tejasranade:
@thomasconner On the naming convention - `storage` is good. `cachePersistence` is misleading (you've noted that it holds more than just the cache) and a little too elaborate. My vote is for `storage`.

On namespace, again given that storage doesn't just apply to `DataStore`, I think it's cleaner to put it outside the `Kinvey.DataStore` context.

I couldn't tell if there were any other open questions outside of these two.

PR #215, opened 2018-01-23
Datastores redesign - Add SQLite persistance 

georgiwe:
#### Description
MLIBZ-2266 Add necessary classes and logic, in order to work with data in SQLite (only in NativeScript).

#### Changes
Add module for working with SQLite. It is used by `SqliteKeyValueStorePersister`. Refactoring during part 4 will take place, where WebSQL logic will be put in a WebSqlModule, justl like `NativescriptSqliteModule`. `SqliteKeyValueStorePersister` will become `SqlKeyValueStorePersister` and it will work with an SQL module - the SQL module will abstract the specifics of executing SQL statements with NativeScript's sqlite plugin, or with the websql interface of the browser. The building of SQL statements and the way they are used will remain in the `SqlKeyValueStorePersister` class. This is to provide maximum code reuse.


src/nativescript/client.ts
```diff
@@ -49,4 +49,8 @@ export class Client extends CoreClient {
     client.activeUserStorage = new ActiveUserStorage();
     return client;
   }
+
+  static sharedInstance() {
```
thomasconner: Why was this added if it just calls `super.sharedInstance()`?

src/core/datastore/repositories/storage-type.js
```diff
@@ -4,5 +4,6 @@ export const storageType = {
   webSql: 'webSql',
   indexedDb: 'indexedDb',
   localStorage: 'localStorage',
-  sessionStorage: 'sessionStorage'
+  sessionStorage: 'sessionStorage',
+  sqlite: 'sqlite'
```
thomasconner: Is there a way to make the storage types specific to the platform being used?

src/nativescript/client.ts
georgiwe: I must have committed this by accident - I was trying out some stuff, doesn't even look like valid syntax. I'll remove it as part of the refactoring ticket. Good catch.

src/core/datastore/repositories/storage-type.js
georgiwe: The enum isn't public yet. I didn't plan on making the enum values the public way of checking what is supported, I thought we'd have an official way, an API (like a method or something). I'd like to reduce the "branching" of structures for each shim. As I said in another comment, this part isn't "officially" implemented so I tried to not go into too much detail. This can be decided as part of the "tweaking" ticket to make the two implementations "meet" :)

PR #217, opened 2018-01-23
Data stores redesign - data persistance backwards compatibility

georgiwe:
#### Description
MLIBZ-2272. Ensure backwards compatibility, when it comes to working with offline data. In this PR, work was done to achieve this in 4 ways:

#### Changes
- improve the persistance selection logic, in order to be able to set defaults for persistance priority lists
- add default priority lists for different platforms, same as the old ones
- change return type of `Kinvey.DataStore.clearCache()` to be `null`, as it was, instead of `undefined`
- remove old classes like `CacheRequest` and `CacheMiddleware` and substitute code using them with code which uses repositories

not exactly backwards compatibility work, but it had to be done
- resolved some circular dependencies and unexpected behaviour in those cases


src/core/client.js
```diff
@@ -142,6 +143,8 @@ export class Client {
      * @private
      */
     this.activeUserStorage = new ActiveUserStorage();
+
+    this.storageType = config.storageType || storageType.inmemory;
```
thomasconner: We just made some changes related to https://kinvey.atlassian.net/browse/MLIBZ-2120 that uses `config.storage` to set the type of storage to use. Is this the same thing that this is doing above? If it is, shouldn't we keep the names the same?

src/core/datastore/datastore.js
```diff
@@ -80,6 +80,7 @@ export class DataStore {
    */
   static clearCache(options = {}) {
     return repositoryProvider.getOfflineRepository()
-      .then(repo => repo.clear());
+      .then(repo => repo.clear())
```
thomasconner: What does `repo.clear()` return? Just curious.

src/core/datastore/sync/utils.js
```diff
@@ -15,9 +16,9 @@ export function groupBy(array, propOrPredicate) {
 
 // TODO: this exists in... datastore?
 const syncOpToPushOpMap = {
-  [RequestMethod.POST]: 'create',
-  [RequestMethod.PUT]: 'update',
-  [RequestMethod.DELETE]: 'delete'
+  [SyncOperation.Create]: 'create',
```
thomasconner: Can you explain why this is needed?

src/core/datastore/datastore.js
georgiwe: It doesn't need to return anything. Right now, I think it returns `true`. But a clear, in my mind, either succeeds, or fails (resolves a promise, or rejects it) - I can't think of a meaningful result for it.

In some later PR, I've set the result of `DataStore.clearCache()` to be `null`, for backwards compatibility.

src/core/client.js
georgiwe: This predates the feature by some time. It will have to conform to the already released API, so a slight change is necessary. I haven't put in too much effort into the implementation, since it wasn't planned to be a big part of the redesign. I just felt we better integrate a basic version of it it early on, to avoid double work later. The specifics will be ironed out now that we've committed to a public API.

src/core/datastore/sync/utils.js
georgiwe: I believe this isn't even used anymore. It'll be removed as part of the refactoring ticket from stage 4.

It used to be a way translate the sync item slang for operations into the SyncOperation slang. Meaning "POST" to "Create". But this is now done differently after a refactoring of SyncManager.

src/core/datastore/datastore.js
thomasconner: Ok. I agree that a clear either succeeds or fails. I just wanted to check to see if it was returning something meaningful that might be helpful. Returning `null` for backwards compatibility makes sense.

src/core/datastore/sync/utils.js
thomasconner: Ok makes sense. Can you point me to where it is now done differently?

src/core/client.js
thomasconner: Ok

src/core/datastore/sync/utils.js
georgiwe: Sure, it's [over here](https://github.com/Kinvey/js-sdk/blob/datastores-redesign-backwards-compatibility/src/core/datastore/sync/sync-manager.js#L113). Essentially having separate methods and a switch.

thomasconner: Ok got it. Thanks

src/core/client.js
thomasconner: This will be fixed with https://kinvey.atlassian.net/browse/MLIBZ-2329.

PR #219, opened 2018-01-23
Data stores redesign - fix push() behaviour

georgiwe:
#### Description
MLIBZ-2310. Update local entities after doing a `.push()`, to ensure any BL changes to the entities are reflected locally, for create and update pushes. And for backwards compatibility.

#### Changes
Add the logic for updating local entities with the responses to create and update network requests (the current server items). Refactor `SyncManager` to improve readability.

### DIFF
--- a/src/core/datastore/sync.spec.js
+++ b/src/core/datastore/sync.spec.js
@@ -1,6 +1,7 @@
 import nock from 'nock';
 import expect from 'expect';
 import chai from 'chai';
+import cloneDeep from 'lodash/cloneDeep';
 import { SyncOperation, syncManagerProvider } from './sync';
 import { SyncStore } from './syncstore';
 import { SyncError } from '../errors';
@@ -253,6 +254,12 @@ describe('Sync', () => {
   });
 
   describe('push()', () => {
+    beforeEach(() => {
+      const store = new SyncStore(collection);
+      return store.clear()
+        .then(() => store.clearSync());
+    });
+
     it('should execute pending sync operations', () => {
       const entity1 = { _id: randomString() };
       const entity2 = { _id: randomString() };
@@ -260,6 +267,12 @@ describe('Sync', () => {
       const entity3Id = randomString();
       const manager = syncManagerProvider.getSyncManager();
       const store = new SyncStore(collection);
+
+      const simulatePreSaveBLHook = (entity) => {
+        const clone = cloneDeep(entity);
+        clone.someNewProperty = true;
+        return clone;
+      };
 
       return store.save(entity1)
         .then(() => store.save(entity2))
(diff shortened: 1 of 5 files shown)


src/core/utils/misc.js
```diff
@@ -29,3 +31,26 @@ export function isValidStorageTypeValue(value) {
   value = ensureArray(value);
   return value.length && value.every(type => supportedPersistances.some(v => type === v));
 }
+
+export function forEachAsync(array, func) {
```
thomasconner: Can you explain how this is different the `Promise.all()`?

test/integration/sync.test.js
```diff
@@ -280,7 +309,7 @@ function testFunc() {
           it('should save the entities from the backend in the cache', (done) => {
             storeToTest.pull()
               .then((result) => {
-                return validatePullOperation(result, [entity1, entity2])
+                return validatePullOperation(result, [entity1, entity2]);
```
thomasconner: I prefer to keep validations in the test itself. Adding functions else where makes it more difficult to follow validations  and require the user to scroll to multiple places. DRY principles are great but I think expectations in unit tests is one area that it doesn't make sense.

Can you put all validations (expectations) inside the unit test?

src/core/utils/misc.js
georgiwe: `Promise.all()` "consolidates" promises into a single promise and resolves when they have all resolved, but rejects as soon as one rejects. This just does a bunch of async operations and lets you know when they've all completed, regardless of their outcome. It's kind of like async.js's `forEach`, but with promises, instead of callbacks.

test/integration/sync.test.js
georgiwe: These are the integration tests.

Having long files anywhere would make you scroll - why does it make less sense in tests, than in source code? I think that's fine. If anything, keeping the assertions stable for all relevant cases is more important in tests - you must know once you've broken something. Keeping them in sync if you have them copied 20 times in a file makes it more risky than not having tests - false positives are worse than no tests as they give you confidence.

I'm inclined to agree that some repetition can be more convenient in tests, but I'd still prefer to avoid it. Like I may be OK with having 3 different suites for CacheStore, SyncStore and NetworkStore, instead of looping over them, cause they have identical CRUD API (though I still lean toward looping, slightly). But the assertions, I see no benefit of repeating the same block of code dozens of times.

In this file, I've added a function called `simulateBLPreSaveHook`. It acts only on the assertions. I made the change in two places. If we had these copied 15 times (the number of tests currently in the sync suite, and it's still not complete), I'd have to do it 15 times.

I think having them this way is fine. But either way, refactoring the integration tests code isn't part of this feature. I've merely added a test to an existing suite.

It's also worth pointing out that these same changes were done outside of the redesign, as it made more sense adding them to master at this point, rather than when we release the redesign. So there was a separate PR for that - the commit after PR can be seen [here](https://github.com/Kinvey/js-sdk/pull/214/commits), as part of the ongoing integration tests effort.

src/core/datastore/sync.spec.js
```diff
@@ -253,6 +254,12 @@ describe('Sync', () => {
   });
 
   describe('push()', () => {
+    beforeEach(() => {
+      const store = new SyncStore(collection);
+      return store.clear()
+        .then(() => store.clearSync());
```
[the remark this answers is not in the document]
georgiwe: Yes, that's the way I have it right now. This must have been from before deciding on the behaviour. It's how I've seen it done elsewhere in the tests code. I guess it's redundant now. I'll remove that as part of the refactoring ticket from stage 4.

test/integration/sync.test.js
thomasconner: Ok

src/core/datastore/sync/sync-manager.js
```diff
@@ -181,20 +192,45 @@ export class SyncManager {
...
-  // TODO: refactor results of individual push ops can be done here?
-  _pushItem({ collection, entityId, state }) {
-    let entity;
+  _handlePushOp(syncItem, offlineEntity) {
+    const { collection, state, entityId } = syncItem;
     const syncOp = state.operation;
 
+    if (!offlineEntity && syncOp !== SyncOperation.Delete) { // todo: duplication
+      const res = this._getPushOpResult(entityId, syncOp);
+      res.error = new KinveyError(`Entity with id ${entityId} not found`);
```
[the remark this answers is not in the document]
georgiwe: What I was going for was backwards compatibility.

Whether we should delete it... I guess we could. I'm usually not comfortable making decisions for the user, but there isn't much more one can do, than deleting it. And it is somewhat cumbersome to find and handle these errors as a user. I think we can delete the sync item.

src/core/utils/misc.js
thomasconner: Got it

src/core/datastore/sync/sync-manager.js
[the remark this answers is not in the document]
tejasranade: Agreed re: backwards compatibility. I realize you were only aiming for feature parity.
These reviews are a good opportunity for me to review our current behavior in more depth - hence the question :)

This particular question stems from recent bug reports where a developer got into a state of "pending push items that fail when pushed". The case being handled in your code here is not exactly that, but the outcome will be the same - the developer is stuck at that point with a corrupt cache.
I agree that we should avoid making decisions on behalf of the developer. In this case, though, the developer is completely out of options. As much as we can, let's keep the state of the cache consistent. 

Feel free to add the improvement here or separately.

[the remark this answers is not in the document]
georgiwe: I think it would be best to add it as part of [MLIBZ-2274](https://kinvey.atlassian.net/browse/MLIBZ-2274), to avoid messing up the chain of PRs now currently open, any further.

I'm adding a comment to the above mentioned ticket about this with a bit of a "spec" :)

PR #220, opened 2018-01-23
MLIBZ-2162: Allow multiple independent datastores on a single collection

thomasconner:
#### Changes
- Added `options.tag` to `Kinvey.DataStore.collection()`. This allows a datastore to interact with a subset of entities in the cache for a collection without providing a query each time.

```javascript
const datastore = Kinvey.DataStore.collection('books', Kinvey.DataStoreType.Cache, { tag: 'Kinvey' });
datastore.find(); // This will find all the books tagged as 'Kinvey' on the books collection
```

#### Tests
- Added a unit test that uses `datastore.pull()` to verify that only a subset of entities are returned for a collection when using a tag

### DIFF
--- a/src/core/datastore/networkstore.js
+++ b/src/core/datastore/networkstore.js
@@ -28,12 +28,12 @@ export class NetworkStore {
       let tag = options.tag;
 
       if (!isString(tag)) {
-        throw new KinveyError('options.tag must be a string.');
+        throw new KinveyError('A tag must be a string.');
       }
 
       tag = tag.trimRight();
-      if (/^[a-zA-Z0-9-]+$/.test(tag)) {
-        throw new KinveyError('options.tag can only contain letters, numbers, and "-".');
+      if (!/^[a-zA-Z0-9-]+$/.test(tag)) {
+        throw new KinveyError('A tag can only contain letters, numbers, and "-".');
       }
 
       this.tag = tag;
--- a/src/core/datastore/networkstore.js
+++ b/src/core/datastore/networkstore.js
@@ -25,6 +25,11 @@ export class NetworkStore {
     this.collection = collection;
 
     /**
+     * @type {string}
+     */
+    this.tag = options.tag || undefined;
+
+    /**
      * @type {Client}
      */
     this.client = options.client;
@@ -107,7 +112,8 @@ export class NetworkStore {
         properties: options.properties,
         query: query,
         timeout: options.timeout,
-        client: this.client
+        client: this.client,
+        tag: this.tag
       };
       let request = new KinveyRequest(config);
 
@@ -156,7 +162,8 @@ export class NetworkStore {
         }),
         properties: options.properties,
         timeout: options.timeout,
-        client: this.client
+        client: this.client,
+        tag: this.tag
       };
       let request = new KinveyRequest(config);
 
--- a/src/core/datastore/networkstore.js
+++ b/src/core/datastore/networkstore.js
@@ -32,8 +32,8 @@ export class NetworkStore {
       }
 
       tag = tag.trimRight();
-      if (tag === '') {
-        throw new KinveyError('options.tag cannot be an empty string.');
+      if (/^[a-zA-Z0-9-]+$/.test(tag)) {
+        throw new KinveyError('options.tag can only contain letters, numbers, and "-".');
       }
 
       this.tag = tag;
--- a/src/core/datastore/networkstore.js
+++ b/src/core/datastore/networkstore.js
@@ -24,6 +24,10 @@ export class NetworkStore {
      */
     this.collection = collection;
 
+    if (options.tag && !isString(options.tag)) {
+      throw new KinveyError('options.tag must be a string.');
+    }
+
     /**
      * @type {string}
      */
--- a/src/core/datastore/networkstore.js
+++ b/src/core/datastore/networkstore.js
@@ -24,14 +24,20 @@ export class NetworkStore {
      */
     this.collection = collection;
 
-    if (options.tag && !isString(options.tag)) {
-      throw new KinveyError('options.tag must be a string.');
-    }
+    if (options.tag) {
+      let tag = options.tag;
 
-    /**
-     * @type {string}
-     */
-    this.tag = options.tag || undefined;
+      if (!isString(tag)) {
+        throw new KinveyError('options.tag must be a string.');
+      }
+
+      tag = tag.trimRight();
+      if (tag === '') {
+        throw new KinveyError('options.tag cannot be an empty string.');
+      }
+
+      this.tag = tag;
+    }
 
     /**
      * @type {Client}
--- a/src/core/datastore/networkstore.js
+++ b/src/core/datastore/networkstore.js
@@ -31,7 +31,6 @@ export class NetworkStore {
         throw new KinveyError('A tag must be a string.');
       }
 
-      tag = tag.trimRight();
       if (!/^[a-zA-Z0-9-]+$/.test(tag)) {
         throw new KinveyError('A tag can only contain letters, numbers, and "-".');
       }
--- a/src/core/datastore/syncstore.spec.js
+++ b/src/core/datastore/syncstore.spec.js
@@ -128,6 +128,39 @@ describe('SyncStore', () => {
         });
     });
 
+    it('should return the entities by tag', () => {
+      const entity1 = { _id: "123foo" };
+      const entity2 = { _id: "123bar" };
+
+      const optionsFoo = { tag: 'foo' };
+      const storeFoo = new SyncStore(collection, optionsFoo);
+      const queryFoo = new Query().equalTo('_id', entity1._id);
+
+      const optionsBar = { tag: 'bar' };
+      const storeBar = new SyncStore(collection, optionsBar);
+      const queryBar = new Query().equalTo('_id', entity2._id);
+
+      nock(storeFoo.client.apiHostname)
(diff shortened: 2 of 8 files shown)


thomasconner:
@heyzooi Is this API similar to how we do it on iOS?

src/core/request/cache.js
```diff
@@ -48,6 +49,18 @@ export class CacheRequest extends Request {
     this._aggregation = aggregation;
   }
 
+  get collection() {
+    if (isDefined(this.tag)) {
+      return  this._collection + this.tag;
```
georgiwe: I think it would be better to separate the collection name and tag by some symbol (or a number of symbols). Preferably a symbol, which cannot be used in a collection name, but can be used in a database name. It would give us some flexibility - we would be able to separate the name from the tag, which can be useful.

thomasconner:
#### Changes
- "tagged" the sync queue operations
- added validation to make sure a tag is a string

src/core/datastore/networkstore.js
```diff
@@ -24,6 +24,15 @@ export class NetworkStore {
      */
     this.collection = collection;
 
+    if (options.tag && !isString(options.tag)) {
+      throw new KinveyError('options.tag must be a string.');
```
georgiwe: I don't think any nonempty string is a valid tag and we should tell the user so, to avoid unnecessary support queries.

thomasconner:
#### Changes
- Added more validation for `options.tag` to make sure it is not an empty string.
- Added unit tests for a tagged sync queue and tag validation

src/core/request/cache.js
thomasconner: I am asking @heyzooi what they did for iOS. We will do the same in the JavaScript SDK.

src/core/datastore/networkstore.js
```diff
@@ -24,6 +24,21 @@ export class NetworkStore {
...
 
+    if (options.tag) {
+      let tag = options.tag;
+
+      if (!isString(tag)) {
+        throw new KinveyError('options.tag must be a string.');
+      }
+
+      tag = tag.trimRight();
+      if (tag === '') {
```
georgiwe: When I said "not every symbol is valid", I meant we should do some research, which symbols are invalid database symbols for the different storage providers and add them to a list. Along with whitespace characters, and check with regex.

And from this, we can address my other comment about the separator between the collection name and the tag in the database collection name, which I feel holds about half the value of validating the tag in the first place. We want to choose a symbol, disallowed for a tag and collection character, but available for a collection name in the database.

I'm sorry if this is getting tedious. We can talk it over before doing the changes, as always. It's just my opinion.

src/core/datastore/syncstore.spec.js
```diff
@@ -128,6 +128,51 @@ describe('SyncStore', () => {
...
+      }).toThrow();
+    });
+
+    it('should throw an error if the tag is an emptry string', () => {
+      expect(() => {
+        new SyncStore(collection, { tag: ' ' });
+      }).toThrow();
+    });
+
+    it('should return the entities by tag', () => {
```
georgiwe: If we're naming this a `spec` file, we should define the full spec in tests. I (as you know) am not a huge fan of unit tests, but we're either doing it, or not. And by this I mean the behaviour for sync-related tasks should be in the spec as well.

thomasconner: Are you referring to the spec for this PR?

src/core/datastore/networkstore.js
heyzooi: @georgiwe let's define what the validation will allow or not

src/core/request/cache.js
thomasconner: @tejasranade What do you think about using a symbol to separate the collection name and the tag? If we ever needed to separate the collection name and tag we could pass these as separate values lower down in the calls.

I can validate the tag against the regex we use for collection names. `^[a-zA-Z0-9-]*$`

src/core/datastore/networkstore.js
thomasconner: I can validate the tag against the regex we use for collection names. `^[a-zA-Z0-9-]*$`

src/core/datastore/syncstore.spec.js
thomasconner: @tejasranade What are your thoughts on this?

src/core/request/cache.js
tejasranade: I was hoping to follow iOS lead on this, but looks like JS will be different.

What to think about - 
- Is it straightforward to identify a tagged collection in the cache vs. a regular one? Think about a debug scenario when someone tells us a table is messed up, for example.
- Can a tagged collection accidentally clash with a regular collection? For example, I use `Project` collection with tag `A`, and have another collection called `ProjectA` (terrible example, I know 😄 )

For lack of a better idea, how about something like `__`? I'm sure there's better options, that's the best I could quickly think of.
I'll leave it to you how to implement it i.e. how it's passed down. It doesn't need to be in the public API obv.

src/core/datastore/syncstore.spec.js
tejasranade: If we're talking about the tests to cover this feature, I do think we need some more tests here. 

My test case was intended to validate our approach. To release this, we should test beyond the base case. In the very least, we should validate a couple of additional typical scenarios - 
- saving data to one of the tagged datastores, followed by a sync.
- clearing data from one of the datastores, and making sure the other isn't cleared 

Both of the above are actually customer expectations for this feature, so I fully expect those to get used.

tejasranade: As a side note - my `foo` / `bar` naming scheme shouldn't make it to master 😄 . We use `randomString()` in all our tests, and let's keep it that way!

src/core/datastore/networkstore.js
georgiwe: That regex would deem an empty string valid - it should be a plus instead of a star, at the end. So I guess "_" is a good candidate for our separateor

src/core/datastore/syncstore.spec.js
georgiwe: I mean in general. Calling these `.spec` files is a TDD convention, no? If that's what we're doing, we should be doing it. It's not just renaming a file. If we're not changing the way we treat those, we might as well leave them as `.test` files.

src/core/request/cache.js
georgiwe: Judging by the regex metioned in a previous comment, it seems like we can go with underscores - they are forbidden for backend collection names. One or more doesn't seem to matter since they are forbidden and can never be part of the actual collection's name. A consideration is collection name length, I guess, and support in different data storage providers (sqlite, websql, indexeddb, different platforms/implementations).

@heyzooi right now, tags create a separate collection in the same database, in JS, so things like empty strings should be rejected, because they would result in the same name (since there is no separator right now). Whitespaces can also be troublesome (tested in websql). For us, the matter is ensuring a subset of characters which are supported in all our storage providers (since we have many, some on more than one platform) and would be useful for customers.

I think alphanumeric characters make sense. Sounds to me like "_" might be part of someone's convention and might be desired for client use.

I can't think of a reason to not create a new database either - why not? Then again, I'm not too familiar with the scenario and what these scoped data stores would be used for outside of this client's use case. Cause we need to be thinking about the general use case, not just this particular one.

src/core/datastore/networkstore.js
georgiwe: Please check my other comment on it. I imagine JS might be the restricting factor in this matter, cause of the numerous data storage providers and platforms. But some research is needed. Only one I feel confident would work everywhere at this point is [A-Za-z] :)

src/core/datastore/syncstore.spec.js
thomasconner: This is out of scope for this PR and should be done in another PR.

src/core/datastore/networkstore.js
thomasconner: I have added the regex `^[a-zA-Z0-9-]+$` for validation. I concatenated the collection name and tag together with a `.`

src/core/datastore/syncstore.spec.js
thomasconner: I added unit tests for saving date to one of the tagged datastore, followed by a push and validating another datastore did not get pushed.

I added a unit tests to clear one tagged datastore and validated that another one did not get cleared.

thomasconner:

src/core/datastore/syncstore.spec.js
georgiwe: I'm not really pushing for a full suite of tests, as much as I'd like us to start developing a specification for a feature before we start working on it. I feel this is a much needed improvement.

src/core/request/cache.js
```diff
@@ -48,6 +49,18 @@ export class CacheRequest extends Request {
     this._aggregation = aggregation;
   }
 
+  get collection() {
+    if (isDefined(this.tag)) {
+      return  `${this._collection}.${this.tag}`;
```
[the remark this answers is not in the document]
thomasconner: Yes it is. We have used it before

src/core/datastore/syncstore.spec.js
tejasranade: @georgiwe if you're referring to TDD, we don't do that as a process today. The spec is supposed to be part of the JIRA ticket description.
Re: what to name the test file, that's a matter of convention and not process, as far as I can see. It's worth following a common convention, but that should be a separate breakout discussion (log it in JIRA? 😄 )

PR #223, opened 2018-01-25
MLIBZ-2031: Cache not returning same items as backend after sync

thomasconner:
#### Changes
- Do not clear the cache when performing a `pull()`.

#### Tests
- Added a unit test that uses multiple pulls and validates the cache.

### DIFF
--- a/src/core/datastore/cachestore.js
+++ b/src/core/datastore/cachestore.js
@@ -862,23 +862,19 @@ export class CacheStore extends NetworkStore {
     options = assign({ useDeltaFetch: this.useDeltaFetch }, options);
     return this.syncManager.pull(query, options)
       .then((entities) => {
-        // Clear the cache
-        return this.clear(query, options)
-          .then(() => {
-            // Save network entities to cache
-            const saveRequest = new CacheRequest({
-              method: RequestMethod.PUT,
-              url: url.format({
-                protocol: this.client.apiProtocol,
-                host: this.client.apiHost,
-                pathname: this.pathname
-              }),
-              properties: options.properties,
-              body: entities,
-              timeout: options.timeout
-            });
-            return saveRequest.execute();
-          })
+        // Save network entities to cache
+        const saveRequest = new CacheRequest({
+          method: RequestMethod.PUT,
+          url: url.format({
+            protocol: this.client.apiProtocol,
+            host: this.client.apiHost,
+            pathname: this.pathname
+          }),
+          properties: options.properties,
+          body: entities,
+          timeout: options.timeout
+        });
+        return saveRequest.execute()
           .then(() => entities);
       });
   }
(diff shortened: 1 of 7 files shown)


thomasconner:
Since we are changing the behavior of `pull()` I am not sure how this might effect things in production apps. When I run the unit tests, all of them still passed after making the change. We might think about releasing this under a beta release.

thomasconner:
@tejasranade Can you review this and approve it if everything looks good? We also should probably discuss this one more tomorrow around what I described above.

thomasconner:
@tejasranade @georgiwe I have made some changes that are ready to review.

src/core/datastore/syncstore.spec.js
```diff
@@ -660,6 +658,7 @@ describe('SyncStore', () => {
         })
         .then((entities) => {
           expect(entities).toEqual([]);
+          return store.clear();
```
georgiwe: What's this about?

src/core/datastore/syncstore.spec.js
```diff
@@ -842,6 +841,42 @@ describe('SyncStore', () => {
           expect(entities).toEqual([entity1, entity2]);
         });
     });
+
+    it('should save the entities from the backend to the cache from multiple pulls', () => {
```
georgiwe: Did this test fail before the fix? Didn't the problem come from the second request finishing before the first? I'm not sure this would happen so with the mocked requests. Perhaps a delay would help and we might as well add a few more entities - 3, 5...

Also shouldn't this be added to the CacheStore suite as well? I think there are more problems there, stemming from the silent "push()". As pointed out in that recent issue.

src/html5/storage/webstorage/localstorage.spec.js
```diff
@@ -0,0 +1,83 @@
+import Promise from 'es6-promise';
```
georgiwe: I think these changes don't pertain to this ticket, do they? About local/session storage and tests. And given the upcoming redesign, these would no longer be relevant.

src/core/datastore/cachestore.js
```diff
@@ -876,26 +876,32 @@ export class CacheStore extends NetworkStore {
...
-              properties: options.properties,
-              body: entities,
-              timeout: options.timeout,
-              tag: this.tag
-            });
-            return saveRequest.execute();
-          })
+        // Remove entities that no longer exist in the backend from the cache
+        const entityIds = entities.map(entity => entity._id);
+        const removeQuery = new Query(query).and().notContainedIn('_id', entityIds);
```
georgiwe: This makes a copy of the user query, so it doesn't change it, right?

src/core/datastore/cachestore.js
```diff
@@ -876,26 +876,32 @@ export class CacheStore extends NetworkStore {
...
+        removeQuery.fields = [];
```
georgiwe: I'm not sure I understand what is going on here - why are we setting these modifiers? Also if we're trying to solve the pagination issues, we'll need a sort order. But I'm fine with focusing on one thing at a time. I'm not sure about the scope here, but I feel I should mention this.

src/core/datastore/cachestore.js
thomasconner: This just makes a copy.

PR #226, opened 2018-01-29
Fixed sort tests after the fix of mlibz-2156

tsvetomir-nedyalkov:
#### Description
After the fix of mlibz-2156 the behaviour is the same offline and online, so the tests should behave accordingly.

### DIFF
--- a/test/integration/crud-entity/common-crud.test.js
+++ b/test/integration/crud-entity/common-crud.test.js
@@ -738,16 +738,14 @@ function testFunc() {
         });
 
         describe('Modifiers', () => {
-          let expectedAscendingCache;
-          let expectedAscendingServer;
+          let expectedAscending;
           let expectedDescending;
 
           describe('Sort', () => {
             before((done) => {
-              expectedAscendingCache = _.sortBy(entities, numberFieldName);
-              expectedAscendingServer = _.sortBy(entities, numberFieldName);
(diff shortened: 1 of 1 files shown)


test/integration/crud-entity/common-crud.test.js
```diff
@@ -738,16 +738,14 @@ function testFunc() {
...
           let expectedDescending;
 
           describe('Sort', () => {
             before((done) => {
-              expectedAscendingCache = _.sortBy(entities, numberFieldName);
-              expectedAscendingServer = _.sortBy(entities, numberFieldName);
-              expectedAscendingServer.splice(0, 0, expectedAscendingServer.pop());
-              expectedDescending = expectedAscendingServer.slice().reverse();
+              expectedAscending = _.sortBy(entities, numberFieldName);
+              expectedAscending.splice(0, 0, expectedAscending.pop());
```
georgiwe: Isn't this just `expectedAscending.unshift(expectedAscending.pop());`? Using splice is a bit confusing. Actually I don't understand this line - is this because `_.sortBy()` puts a `null` or `undefined` value in the end, and the server puts it at the front? I think a small comment about it would be nice, since it's not obvious.

tsvetomir-nedyalkov: Yes, the reason is that the server sorts null values at the beginning - changed the splice with unshift.

PR #227, opened 2018-02-02
Datastores redesign stage 4, part 1 - refactor redesign code

georgiwe:
#### Description
MLIBZ-2273. Refactor parts of the redesign which I had noted as in need of attention.

#### Changes
Removed old (now unused) files, imports and function arguments. Extract specific SQL implementations into SQL modules, used by an SqlKeyValueStorePersister. Some readability improvements in SyncManager and tests. Remove outdated TODOs.


src/core/datastore/cachestore.js
```diff
@@ -81,14 +76,14 @@ export class CacheStore extends NetworkStore {
    *                                                                            from the local cache.
    * @return  {Promise}                                                         Promise
    */
-  pendingSyncCount(query, options) {
+  pendingSyncCount(query) {
```
thomasconner: You should remove the information about the `options` in the documentation for the function.

src/core/datastore/datastore.js
```diff
@@ -78,7 +78,7 @@ export class DataStore {
    * @param  {Object} [options={}] Options
    * @return {Promise<Object>} The result of clearing the cache.
    */
-  static clearCache(options = {}) {
+  static clearCache() {
```
thomasconner: You should remove the information about the `options` in the documentation for the function.

src/core/datastore/cachestore.js
georgiwe: If there's nothing more to change on this PR, can we do this in a later PR? I'd like to avoid rebasing 3 branches (and complicating the review of the other two open PRs) because of removing a comment. I'll do it in part 4 (the one I'm working on right now), how does that sound?

thomasconner: That is fine with me.

PR #231, opened 2018-02-06
Integration tests additions

georgiwe:
#### Description
Add a few integration tests. Change some tests to use the Observable API, instead of promises, to make sure we have that covered on all methods of data stores. Group some tests in a describe block, instead of featuring the method being tested in their names.

Some tests are skipped, as they are about functionality which is not yet implemented.

### DIFF
--- a/test/integration/tests/crud-entity/common-crud.test.js
+++ b/test/integration/tests/crud-entity/common-crud.test.js
@@ -153,9 +153,13 @@ function testFunc() {
 
             storeToTest.findById(entityId).subscribe(nextHandlerSpy, (err) => {
               const expectedCallCount = dataStoreType === Kinvey.DataStoreType.Cache ? 1 : 0;
-              expect(nextHandlerSpy.callCount).to.equal(expectedCallCount);
-              expect(err.name).to.contain(notFoundErrorName);
-              done();
+              try {
+                expect(nextHandlerSpy.callCount).to.equal(expectedCallCount);
+                expect(err.name).to.contain(notFoundErrorName);
+              } catch (err) {
+                return done(err);
+              }
+              return done();
             }, () => {
               done(new Error('Should not be called'));
             });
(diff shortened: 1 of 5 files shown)


test/integration/tests/crud-entity/common-crud.test.js
```diff
@@ -149,22 +149,26 @@ function testFunc() {
...
           it('should throw a NotFoundError if the id argument does not exist', (done) => {
             const entityId = utilities.randomString();
-            storeToTest.findById(entityId).toPromise()
-              .then(() => done(new Error('Should not be called')))
-              .catch((error) => {
-                expect(error.name).to.contain(notFoundErrorName);
-                done();
-              })
-              .catch(done);
+            const nextHandlerSpy = sinon.spy();
```
tsvetomir-nedyalkov: It seems that the tests timeouts if the assertion fails

tsvetomir-nedyalkov: The test ''should throw a NotFoundError if the id argument does not exist''

test/integration/tests/crud-entity/offline-crud.test.js
```diff
@@ -63,6 +63,25 @@ function testFunc() {
               .catch(done);
           });
 
+          it('findById() should create in the cache, an entity found on the backend, but missing in the cache', (done) => {
```
tsvetomir-nedyalkov: It seems that the test does not fail if the verification utilities.validateEntity(dataStoreType, collectionName, entity); fails.

test/integration/tests/crud-entity/common-crud.test.js
georgiwe: Wrapped in a try/catch to avoid timing out.

test/integration/tests/crud-entity/offline-crud.test.js
georgiwe: I hadn't noticed it was an async method. Fixed.

PR #233, opened 2018-02-06
MLIBZ-2159: _id field not returned when specifying fields for a query

thomasconner:
#### Changes
- Added `const PROTECTED_FIELDS = ['_id', '_kmd', '_acl'];`
- Concatenated the specified fields array with the protected fields array

#### Tests
- Added unit test to verify protected fields are not removed when using a query with specified fields

### DIFF
--- a/src/core/query.js
+++ b/src/core/query.js
@@ -4,7 +4,7 @@ import { QueryError } from './errors';
 import { nested, isDefined, isNumber } from './utils';
 import { Log } from './log';
 
-const PROTECTED_FIELDS = ['_id', '_kmd', '_acl'];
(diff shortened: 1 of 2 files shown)


src/core/query.js
```diff
@@ -4,6 +4,7 @@ import { QueryError } from './errors';
 import { nested, isDefined, isNumber } from './utils';
 import { Log } from './log';
 
+const PROTECTED_FIELDS = ['_id', '_kmd', '_acl'];
```
georgiwe: Why have `_kmd` in there too - it's not returned by the server.

PR #234, opened 2018-02-07
Datastores redesign stage 4, part 3 - storage selection

georgiwe:
#### Description
MLIBZ-2329. Change the selection of storage to fit the implementation in master currently. There is also a slight change to an integration test about `clear()`. I separated testing deletion of collection entities from deletion of sync items. This is one of the changes with the current implementation that are coming with the redesign.

#### Changes
Rename the storage enum, expose it in the same way (with only supported storages for the platform).

### DIFF
--- a/src/core/utils/misc.js
+++ b/src/core/utils/misc.js
@@ -28,8 +28,8 @@ export function ensureArray(obj) {
 
 export function isValidStorageProviderValue(value) {
   const supportedPersistances = repositoryProvider.getSupportedStorages();
-  value = ensureArray(value);
-  return value.length && value.every(type => supportedPersistances.some(v => type === v));
+  const valueAsArray = ensureArray(value);
+  return !!value && valueAsArray.length && valueAsArray.every(type => supportedPersistances.some(v => type === v));
 }
 
 export function forEachAsync(array, func) {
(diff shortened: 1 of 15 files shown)


src/core/utils/misc.js
```diff
@@ -54,3 +54,10 @@ export function forEachAsync(array, func) {
     });
   });
 }
+
+export function useIfDefined(value, defaultValue) {
+  if (value !== undefined) {
```
thomasconner: This should use `if (typeof value !== 'undefined'`). `undefined` can be redefined in the global namespace in JavaScript.

georgiwe: I'm not sure this is allowed nowadays. And if anyone is doing that, they have bigger problems than this check not working out ;) If there are any other changes required, I can make this one. But it's not worth the hassle on its own, imo.

thomasconner: Ok so it seems that `undefined` is no longer writable in the global namespace in JavaScript since 1.8.5 according to MDN. However, this is definitely possible nowadays using a self executing function.  Consider the two following code snippets.

```javascript
(function() { 
  var undefined = 'foo'; 
  console.log(undefined, typeof undefined);
})();
```

```javascript
(function(undefined) { 
  console.log(undefined, typeof undefined); 
})('foo');
```

Both code snippets will print `foo string` to the console. For this reason we should use the `typeof` way of checking if a value is `undefined`.

georgiwe: Yes, it is possible to name a variable "undefined". I honestly think the chances of this happening here are practically 0. But I made the change and will have Tsvetomir start over with testing tomorrow.

PR #238, opened 2018-02-12
Datastores redesign stage 4, part 4 - tagged datastores

georgiwe:
#### Description
MLIBZ-2338 and MLIBZ-2346. I figured MLIBZ-2346 isn't large enough to require a PR in itself.

#### Changes
Implement tagged data store feature. Stop skipping tests for it, add a few more unit tests. Also fixed the order of delete and create operations when trying to update the local result of a create() with the network result (CacheStore only).

### DIFF
--- a/src/nativescript/push/common.ts
+++ b/src/nativescript/push/common.ts
@@ -160,6 +160,7 @@ export class PushCommon extends EventEmitter {
     }
 
     const device = {
+      _id: activeUser._id,
       userId: activeUser._id,
       token: token
     };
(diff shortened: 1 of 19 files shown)


src/phonegap/push.js
```diff
@@ -190,7 +190,7 @@ export class PushNotification extends EventEmitter {
           })
           .then((response) => {
             if (response) {
-              return response.data;
+              return response;
```
georgiwe: This is part of MLIBZ-2374, but I figured it didn't warrant a PR in itself.

src/nativescript/push/common.ts
```diff
@@ -160,6 +160,7 @@ export class PushCommon extends EventEmitter {
     }
 
     const device = {
+      _id: activeUser._id,
```
georgiwe: Fix for MLIBZ-2376.

PR #241, opened 2018-02-15
MLIBZ-2350 Revert removal of caching db connections in WebSQL.

georgiwe:
#### Description
During the mono repo move (I believe. this is the commit [here](https://github.com/kinvey/js-sdk/commit/73903fa)), the caching of database connections was removed. This caused problems when making many operations in websql, since each one would "open" the database. We encountered this in the unit tests suite.

#### Changes
Add back the caching of "open" databases.

### DIFF
--- a/src/html5/storage/websql.js
+++ b/src/html5/storage/websql.js
@@ -5,6 +5,7 @@ import { isDefined } from '../../core/utils';
 const masterCollectionName = 'sqlite_master';
 const size = 2 * 1024 * 1024; // Database size in bytes
 let isSupported;
+const dbCache = {};
 
 export class WebSQL {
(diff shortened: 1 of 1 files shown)


georgiwe:
@thomasconner , why was this removed in the first place? Perhaps we shouldn't bring it back? Also, there didn't seem to be a mechanism for closing or removing these from cache. Is that the way it should be?

@tsvetomir-nedyalkov asked me to take this on, since it's causing problems with integration tests and would force him to disable running the tests on phonegap in general. I had tracked it down to this perticular commit, so I figured it won't be too difficult to revert the change. And if this is all that needs doing, we better do it sooner, rather than later, and have a more robust integration test suite.

PR #242, opened 2018-02-15
MLIBZ-2314 Updated typescript definitions

tejasranade:
#### Description
Updated definition of `signup()` to allow zero parameters (required to signup as implicit user)

#### Changes
Minor change to `.ts` file.

### DIFF
--- a/src/kinvey.d.ts
+++ b/src/kinvey.d.ts
@@ -26,6 +26,7 @@ export namespace Kinvey {
     timeout ? : number;
     useDeltaFetch ? : boolean;
     version ? : string;
+    micId? : string;
   }
 
   // ClientConfig interface
@@ -545,6 +546,7 @@ interface RequestOptions {
   timeout?: number;
   useDeltaFetch?: boolean;
   version?: string;
+  micId?: string;
 }
 
 // ClientConfig interface
--- a/src/kinvey.d.ts
+++ b/src/kinvey.d.ts
@@ -544,7 +544,7 @@ interface RequestOptions {
   properties?: Properties;
   timeout?: number;
   useDeltaFetch?: boolean;
-  version?: String;
+  version?: string;
 }
 
 // ClientConfig interface
--- a/src/kinvey.d.ts
+++ b/src/kinvey.d.ts
@@ -436,8 +436,8 @@ export namespace Kinvey {
     static loginWithMIC(redirectUri: string, authorizationGrant ? : AuthorizationGrant, options ? : RequestOptions): Promise < User > ;
     logout(options ? : RequestOptions): Promise < void > ;
     static logout(options ? : RequestOptions): Promise < void > ;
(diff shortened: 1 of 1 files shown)


src/kinvey.d.ts
```diff
@@ -544,6 +544,7 @@ interface RequestOptions {
   properties?: Properties;
   timeout?: number;
   useDeltaFetch?: boolean;
+  version?: String;
```
tejasranade: This change should've been in my previous typescript PR #235 . I did not realize we had two copies of the ts definitions in the same file.

src/kinvey.d.ts
```diff
@@ -544,6 +544,7 @@ interface RequestOptions {
   properties?: Properties;
   timeout?: number;
   useDeltaFetch?: boolean;
+  version?: string;
```
tejasranade: This change should've been in my previous typescript PR #235 . I did not realize we had two copies of the ts definitions in the same file.

src/kinvey.d.ts
```diff
@@ -436,8 +436,8 @@ export namespace Kinvey {
     static loginWithMIC(redirectUri: string, authorizationGrant ? : AuthorizationGrant, options ? : RequestOptions): Promise < User > ;
     logout(options ? : RequestOptions): Promise < void > ;
     static logout(options ? : RequestOptions): Promise < void > ;
-    signup(data: {}, options ? : RequestOptions): Promise < this > ;
-    static signup(data: {}, options ? : RequestOptions): Promise < User > ;
+    signup(data ? : {}, options ? : RequestOptions): Promise < this > ;
```
georgiwe: Shouldn't the data type of all of these be "any"?

tejasranade: I had to look this up. Turns out, there's a subtle difference - https://stackoverflow.com/questions/18961203/typescript-any-vs-object

Based on that article (see last note in this answer - https://stackoverflow.com/a/28795689), I feel using `object` is the most appropriate. We want to be as restrictive as possible, to avoid bad input. Thoughts?

georgiwe: I guess when defining a method argument in a definition file, we can use `object` and it is better. Although it still allows you to pass a function.

But don't we know even more about this object? Like it should have a `username` and `password` properties, no?

thomasconner: I think we do but it can also have much more then just these properties. Lets look at updating these definitions going forward but lets not hold up this change since currently it will throw and error if no `data` is provided.

georgiwe: Sure, I wasn't requesting changes at any point.

But the way to define the data type is something like this:

```typescript
interface UserData {
  [key: string]: value: any
}

interface UserSignUpData extends UserData {
  username: string,
  password: string
}
```

This way you have the two properties, but can add as many more as you like.

PR #245, opened 2018-02-20
(Query -> isSupportedOffline) Check that a value is an object before calling Object.keys

chrismllr:
#### Description
When utilizing a `Cache` collection, requests are failing in this function block due to the `value` being either a string or a boolean, and are being passed into `Object.keys`.

For some reference, when this issue was encountered, the value for `this.filter` was:
```js
{ collection: "GenericUser" }
```

Which is the name of the Cache collection we are calling `find` on. `value`, in this case, will equal `"GenericUser"`, which is breaking when being passed to `Object.keys`.

I haven't done too much digging to see whether this was an issue with how `CacheRequest` is passing the current query, or if this is just an edge case that should be included in this function.. any insight would be great!

#### Changes
- Add a type check to make sure `value` is an object before calling `Object.keys` on it.

### DIFF
--- a/src/core/query.js
+++ b/src/core/query.js
@@ -197,7 +197,11 @@ export class Query {
       if (supported) {
         const value = this.filter[key];
         return UNSUPPORTED_CONDITIONS.some((unsupportedConditions) => {
-          if (!value || typeof value !== 'object') {
+          if (!value) {
+            return true;
+          }
+
+          if (typeof value !== 'object') {
             return true;
           }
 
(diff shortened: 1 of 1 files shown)


thomasconner:
@chrismllr Thanks for the PR! Can you add some unit tests for this change in [query.spec.js](https://github.com/Kinvey/js-sdk/blob/master/src/core/query.spec.js#L112)? 

To run the complete suite of unit tests just execute `npm test` at the root of the project. You can also run a subset of the complete suite of unit tests by executing `npm test -- --grep "isSupportedOffline()"`.

PR #247, opened 2018-02-21
MLIBZ-2316: Server Side Delta Set

thomasconner:
Add new server side delta set functionality.

#### Changes
- Updated `request/deltaFetch.js` to perform server side delta set requests.

#### Tests
- Verify `request/deltaFetch.js` sends correct requests.
- Verify `datastore/cachestore.js` behaves correctly when `useDeltaFetch` is turned on.


thomasconner:
Sorry about the committer of the commits being shown as TravisCI. I changed my `user.name` and `user.email` after I realized this issue so going forward it will show my name.

src/core/datastore/sync/sync-manager.js
```diff
@@ -60,8 +62,31 @@ export class SyncManager {
     if (isEmpty(collection)) {
       return Promise.reject(new KinveyError('Invalid or missing collection name'));
     }
-    return this._fetchItemsFromServer(collection, query, options)
-      .then(entities => this._replaceOfflineEntities(collection, query, entities));
+
+    return this.getSyncItemCountByEntityQuery(collection, query)
```
thomasconner: @georgiwe Not sure if this is being done somewhere else but please verify that we are not duplicating this check. I had to add this here to fix a failing unit tests. All unit tests pass now and I wasn't able to find this check anywhere else after doing a search.

src/core/datastore/cachestore.spec.js
```diff
@@ -190,51 +190,87 @@ describe('CacheStore', () => {
...
+                expect(onNextSpy.calls[1].arguments).toEqual([[entity2]]);
+                done();
               } catch (error) {
                 done(error);
               }
             });
         });
     });
+
+    // it('should remove entities that no longer exist on the backend from the cache', (done) => {
```
thomasconner: @georgiwe @tejasranade Lets discuss this unit test after standup one day.

src/core/datastore/processors/cache-offline-data-processor.js
```diff
@@ -79,16 +79,23 @@ export class CacheOfflineDataProcessor extends OfflineDataProcessor {
...
-          return this._replaceOfflineEntities(collection, offlineEntities, networkEntities);
+        .then(() => {
+          return super._processRead(collection, query, options);
+        })
+        .then((entities) => {
+          observer.next(entities);
+          observer.complete();
+          return entities;
+        })
+        .catch((error) => {
```
georgiwe: This isn't necessary - the `wrapInObservable` method should handle it.

src/core/datastore/processors/cache-offline-data-processor.js
```diff
@@ -79,16 +79,23 @@ export class CacheOfflineDataProcessor extends OfflineDataProcessor {
   _processRead(collection, query, options) {
     let offlineEntities;
     return wrapInObservable((observer) => {
-      return this._ensureCountBeforeRead(collection, 'fetch the entities', query)
```
georgiwe: This is a change in behaviour, no?

src/core/datastore/processors/cache-offline-data-processor.js
```diff
@@ -79,16 +79,23 @@ export class CacheOfflineDataProcessor extends OfflineDataProcessor {
...
           offlineEntities = entities || []; // backwards compatibility
           observer.next(offlineEntities);
-          return this._networkRepository.read(collection, query, options);
+          return this._syncManager.pull(collection, query, options);
         })
-        .then((networkEntities) => {
-          observer.next(networkEntities);
-          return this._replaceOfflineEntities(collection, offlineEntities, networkEntities);
+        .then(() => {
+          return super._processRead(collection, query, options);
```
georgiwe: Reading them again? I think you mentioned something about this yesterday, but could you elaborate?

src/core/datastore/sync/sync-manager.js
georgiwe: Which test, why was it failing? The reason this was in CacheStore.pull() was that it's logic that isn't part of `pull()`. I like to keep concerns separate. `SyncManager.pull()` does work, CacheStore.pull() should make sure "business logic" like this is handled before calling it to do the work.

src/core/datastore/sync/sync-manager.js
```diff
@@ -269,9 +294,29 @@ export class SyncManager {
...
-  _fetchItemsFromServer(collection, query, options) {
-    // TODO: deltaset logic goes here
-    return this._networkRepo.read(collection, query, options);
+  _fetchItemsFromServer(collection, query, options = {}) {
+    return this._networkRepo.read(collection, query, options)
+      .then((data) => {
+        const useDeltaSet = options.useDeltaFetch || false;
+
+        if (useDeltaSet) {
+          return this._getOfflineRepo()
```
georgiwe: This warrants its own method.

src/core/datastore/sync/sync-manager.js
```diff
@@ -269,9 +294,29 @@ export class SyncManager {
...
+            .then((repo) => {
+              const { deleted } = data;
+
+              if (isArray(deleted) && deleted.length > 0) {
+                const deletedIds = deleted.map((item) => item._id);
+                const deleteQuery = new Query().containsAll('_id', deletedIds);
```
georgiwe: I believe a `$all` filter wouldn't work the same as `$in`, would it? This would never match anything, would it?

src/core/datastore/sync/sync-manager.js
```diff
@@ -60,8 +62,31 @@ export class SyncManager {
...
+        // Throw an error if there are still items that need to be synced
+        if (count > 0) {
+          return Promise.reject(
+            new SyncError('Unable to fetch the entities on the backend.'
+              + ` There are ${count} entities that need to be synced.`)
+          );
+        }
+
+        return this._fetchItemsFromServer(collection, query, options)
+          .then(entities => this._replaceOfflineEntities(collection, query, entities));
```
georgiwe: After seeing the changed `_fetchItemsFromServer()`, `_replaceOfflineEntities()` has to change too - it doesn't need to delete anything. If `_fetchItemsFromServer()` worked as expected, it should have deleted everything that needs deleting. Currently, it has the wrong filter, so this delete inside `_replaceOfflineEntities()` is the one that is doing the actual work.

src/core/request/deltafetch.js
```diff
@@ -190,4 +122,26 @@ export class DeltaFetchRequest extends KinveyRequest {
 
     return this.url.substring(ind + appkeyStr.length, indOfQueryString);
   }
+
+  static execute(options, client, dataOnly = true) {
```
georgiwe: Isn't this inherited?

src/core/request/deltafetch.js
```diff
@@ -46,132 +42,68 @@ export class DeltaFetchRequest extends KinveyRequest {
 
   execute() {
     return repositoryProvider.getOfflineRepository()
-      .then(repo => repo.read(this._getCollectionFromUrl(), this.query))
-      .catch((error) => {
-        if (!(error instanceof NotFoundError)) {
-          throw error;
+      .then((repo) => {
+        const collectionName = this._getCollectionFromUrl();
+        let queryCacheQuery = new Query()
```
georgiwe: There are a couple of things I am not clear on here:

- tagged datastores - is it clear how they would work in this regard?
- the serialization of queries and general contents of the "query cache" collection

src/core/request/deltafetch.js
```diff
@@ -46,132 +42,68 @@ export class DeltaFetchRequest extends KinveyRequest {
...
+          .equalTo('collectionName', collectionName)
+          .descending('lastRequest');
+
+        if (this.query) {
+          queryCacheQuery = queryCacheQuery
+            .and()
+            .equalTo('query', this.query.toString());
```
georgiwe: Wouldn't this include `skip` and `limit`? I thought it was decided these will not be part of the persisted query representation. What about sort - should that be part of it? Or maybe just filter, cause I don't think the order of the returned entities matters for our purposes, does it?

src/core/request/deltafetch.js
```diff
@@ -46,132 +42,68 @@ export class DeltaFetchRequest extends KinveyRequest {
...
+              return request.execute();
+            }
+
+            return request.execute();
+          })
+          .then((response) => {
+            // Makse sure the response from DeltaSet is normalized
+            const { data } = response;
+
+            if (!Object.prototype.hasOwnProperty.call(data, 'changed')) {
```
georgiwe: Surely the server doesn't return objects, which have inherited properties - can't we leave out the ugliness of `hasOwnProperty`?

src/core/request/deltafetch.js
```diff
@@ -46,132 +42,68 @@ export class DeltaFetchRequest extends KinveyRequest {
...
+
+            if (!Object.prototype.hasOwnProperty.call(data, 'deleted')) {
+              data.deleted = [];
+            }
+
+            response.data = data;
+            return response;
+          })
+          .then((response) => {
+            return repo.create(QUERY_CACHE_COLLECTION_NAME, {
```
georgiwe: Shouldn't this be an `update()`? Which is actually an upsert, which is what we are trying to do, no?

src/core/datastore/processors/cache-offline-data-processor.js
thomasconner: I am doing a read operation on the cache again because this was the only reliable way to get the full set of entities that match the original call after a delta set has been performed.

src/core/datastore/processors/cache-offline-data-processor.js
```diff
@@ -79,16 +79,23 @@ export class CacheOfflineDataProcessor extends OfflineDataProcessor {
...
+        })
+        .then((entities) => {
```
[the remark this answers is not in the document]
thomasconner: Ok I wasn't aware. I will remove it.

src/core/datastore/sync/sync-manager.js
thomasconner: Yes I meant to use a `$in` filter.

src/core/request/deltafetch.js
thomasconner: Yes I should be removing `skip` and `limit`. I am not sure about `sort`. I will just use `filter` for now.

src/core/request/deltafetch.js
thomasconner: Yes I can

src/core/request/deltafetch.js
thomasconner: Yes it should be an update.

src/core/request/deltafetch.js
thomasconner: It is but the inherited function will do `new KinveyRequest()` and not `new DeltaFetchRequest()`.

src/core/request/deltafetch.js
thomasconner: I need to add a unit test for using a tagged datastore with delta fetch and make changes if things don't behave as expected. Good catch.

src/core/datastore/sync/sync-manager.js
thomasconner: I was referring to [this unit test](https://github.com/Kinvey/js-sdk/pull/247/files#diff-50ea438538b5fab8b06c67a1ba2639caR230)

thomasconner: I am seeing some weird behavior that is causing the unit tests [here](https://github.com/Kinvey/js-sdk/pull/247/files#diff-50ea438538b5fab8b06c67a1ba2639caR114) and [here](https://github.com/Kinvey/js-sdk/pull/247/files#diff-50ea438538b5fab8b06c67a1ba2639caR230) to fail however I have debugged and the logic appears to be correct. Maybe another pair of eyes on the unit tests will help see something I am missing.

src/core/datastore/processors/cache-offline-data-processor.js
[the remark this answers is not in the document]
thomasconner: Ok I changed it.

src/core/datastore/cachestore.spec.js
```diff
@@ -111,130 +111,206 @@ describe('CacheStore', () => {
         });
     });
 
-    it('should return the entities', (done) => {
+    it('should return the entities');
```
georgiwe: It's very difficult to read the diff when tests are being changed this way. The best way to mark it as pending is to call `skip`.

Like so
`it.skip('should return the entities', () => {...})`.

The lines below are "out of sync" - they don't appear next to where they were changed, due to identical lines in different tests. So the entire diff is pretty hard to read. Doing it as mentioned above would make things much better.

thomasconner: Ok I wasn't aware of this options.

PR #252, opened 2018-02-23
MLIBZ-2385 Incorrect sync item count

georgiwe:
#### Description
Querying for deleted entities is still a bit unclear ([MLIBZ-2177](https://kinvey.atlassian.net/browse/MLIBZ-2177)), so the special case where no query is provided needs to be handled explicitly and this was missing from the `pull()` method.

Also added a tiny optimization for the type of query being built for sync items. This has to do with [the difference](https://hackmd.io/IbBmGMAYFYGYBMC0B2ARgFgIyPeUBORVaVZRWWXLdfTcWZIA) that the collection name of sync items now contains the tag. So the filter needs to match the tagged or untagged collection name (for backwards compatibility), so it has an "or" condition. But when the collection being used doesn't have a tag in the first place, it would result in a redundant "or", so I removed it to speed up inmemory filtering. This can be removed in a future version, since it's unlikely sync items haven't been synced for that long, so all new sync items will have the new collection name.

#### Changes
Use different method for sync entity count, depending on the presence of a query. Only add "or" filter when using a tagged collection.

### DIFF
--- a/src/core/datastore/sync/sync-manager.js
+++ b/src/core/datastore/sync/sync-manager.js
@@ -74,6 +74,10 @@ export class SyncManager {
 
   // TODO: this method is temporray, pending fix for MLIBZ-2177
   getSyncItemCountByEntityQuery(collection, query) {
+    if (!query) {
+      return this._syncStateManager.getSyncItemCount(collection);
+    }
+
     return this._getOfflineRepo()
       .then(repo => repo.read(collection, query))
       .then((entities) => {
@@ -82,8 +86,8 @@ export class SyncManager {
       });
   }
 
+  // TODO: this only returns nondeleted entities - pending fix for MLIBZ-2177
   getSyncEntities(collection, query) {
-    // TODO: this only returns nondeleted entities - pending fix for MLIBZ-2177
     return this._getOfflineRepo()
       .then(repo => repo.read(collection, query))
       .then((entities = []) => {
--- a/src/core/datastore/sync/sync-state-manager.js
+++ b/src/core/datastore/sync/sync-state-manager.js
@@ -13,7 +13,8 @@ import {
   generateEntityId,
   getTagFromCollectionName,
   formTaggedCollectionName,
-  stripTagFromCollectionName
+  stripTagFromCollectionName,
+  collectionHasTag
 } from '../utils';
 
 // imported for typings
@@ -123,8 +124,10 @@ export class SyncStateManager {
   _getCollectionFilter(collection) {
     const result = new Query();
     result.equalTo('collection', collection)
-      .or()
-      .equalTo('collection', stripTagFromCollectionName(collection));
+    if (collectionHasTag(collection)) {
+      result.or()
+        .equalTo('collection', stripTagFromCollectionName(collection));
+    }
     return result;
   }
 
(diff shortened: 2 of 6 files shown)


src/core/datastore/sync/sync-manager.js
```diff
@@ -74,6 +74,10 @@ export class SyncManager {
 
   // TODO: this method is temporray, pending fix for MLIBZ-2177
   getSyncItemCountByEntityQuery(collection, query) {
+    if (!query) {
+      return this._syncStateManager.getSyncItemCount(collection);
```
thomasconner: Why not call `this.getSyncItemCount(collection)`?

src/core/datastore/sync/sync-state-manager.js
```diff
@@ -123,8 +124,10 @@ export class SyncStateManager {
   _getCollectionFilter(collection) {
     const result = new Query();
     result.equalTo('collection', collection)
-      .or()
-      .equalTo('collection', stripTagFromCollectionName(collection));
+    if (collectionHasTag(collection)) {
+      result.or()
+        .equalTo('collection', stripTagFromCollectionName(collection));
```
thomasconner: Can you explain how this `equalTo` clause would ever match? It seems like it would never match a collection name.

georgiwe: The variable `collection`'s value can be either "books" or "books.my-tag". Before the redesign, the value of any sync item's `.collection` property used to be the untagged collection. After the redesign, it's the tagged collection (if it has a tag). So for sync items created before the redesign and still in the queue after an update to the redesigned version, the $or clause must be added - for backwards compatibility. After some time, this can be removed, the assumption being that whatever needed to be synced, was already synced and any new sync items only have the "new" way of setting the value (with the tag).

As to how it would match, it would add an $or clause, if the value of `collection` has a tag. So two cases:

- `collection === 'books'` - no $or statement, just `{ collection: 'books' }`
- `collection === 'books.my-tag'` - $or statement, resulting filter: `{ $or: [{ collection: 'books' }, { collection: 'books.my-tag' }] }`

Am I missing something in the `Query` API? Cause I had a hard time figuring out when to chain calls and when not to. But it appears to work as expected. Which part do you think would fail?

thomasconner: It might be that I don't understand what is returned from `stripTagFromCollectionName(collection)`. I am thinking the query filter would be: `{ $or: [{ collection: 'books' }, { collection: 'my-tag' }] }`.

thomasconner: Ok this makes sense now to me.

src/core/datastore/sync/sync-manager.js
georgiwe: `getSyncItemCount` doesn't accept a query, since it can't reliably apply an entity query to the sync queue - the only info about the entities themselves it has is their id. So it accepts a list of entity ids, for which to return the sync items (the only type of entity querying available). So if you pass a query, you have to do a read() with it first and use the ids of matching entities. So that's what's going on.

thomasconner: Ok got it

PR #256, opened 2018-02-26
Auto pagination

georgiwe:
#### Description
MLIBZ-2045. Auto pagination for pull() method of data stores.

This is a preliminary version, PR is for discussion purposes.

Currently, the auto pagination setting is one of the options for the pull() method.

```javascript
syncStore.pull(query, { autoPagination: true });
```

It can also be an object, specifying more settings:

```javascript
const paginationOptions = { pageSize: 1000, pullLimit: 23000 };
const options = { autoPagination: paginationOptions };
syncStore.pull(query, options);
```

The `pullLimit` setting is something I wanted to try out. It was meant for cases when one might want to use auto pagination, but not pull all entities, matching the query. It was part of what we discussed today, during standup. What are your thoughts on that?

Ping @tejasranade @thomasconner @vinaygahlawat @heyzooi


src/core/datastore/sync/sync-manager.js
```diff
@@ -347,4 +352,66 @@ export class SyncManager {
...
+          return this._fetchAndUpdateEntities(collection, query);
+        });
+      });
+  }
+
+  _paginate(collection, userQuery, options) {
+    const query = new Query({ filter: userQuery.filter }); // ignore sort, it's irrelevant
+    return this._networkRepo.count(collection, query)
+      .then((totalCount) => {
+        const pullLimit = this._getPullLimit(options);
```
tejasranade: I haven't understood this parameter. What's `pullLimit` supposed to signify? Typically, developers will limit their results by providing a query, not a numerical limit.

src/core/datastore/sync/sync-manager.js
```diff
@@ -347,4 +352,66 @@ export class SyncManager {
         return Promise.reject(new SyncError('Invalid sync event name'));
     }
   }
+
+  _getPullLimit(pullOptions) {
+    const limitOption = pullOptions && pullOptions.pullLimit;
+    return limitOption || Infinity;
+  }
+
+  _getInternalPullQuery({ filter, sort, fields }, pullLimit) {
```
tejasranade: I don't understand the significance of this method. We're going to call `_getPaginationQueries` anyway, so creating this `pullQuery` with meaningless skip and limit seems unnecessary.
Nothing wrong with the method per se, but it'll be cleaner to just create all the pagination queries in one place.

src/core/datastore/sync/sync-manager.js
```diff
@@ -347,4 +352,66 @@ export class SyncManager {
...
+      .then(() => {
+        // TODO: throttle max request count - reuse from push?
+        return forEachAsync(queries, (query) => {
+          return this._fetchAndUpdateEntities(collection, query);
+        });
+      });
+  }
+
+  _paginate(collection, userQuery, options) {
+    const query = new Query({ filter: userQuery.filter }); // ignore sort, it's irrelevant
```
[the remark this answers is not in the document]
georgiwe: It's only for the _count query. Other queries respect filter, sort and fields.

Perhaps it adds a tiny bit to the learning of how auto pagination works, but I don't think anyone would care much if their sort is missing from the _count request, especially since it doesn't make sense there, or maybe even causes a slowdown performance-wise (I'm not sure how backend treats this sort of query, or Mongo, but still).

That said, I don't feel very strongly either way - I will add back fields and sort, if you feel they are useful.

src/core/datastore/sync/sync-manager.js
georgiwe: I mentioned it in the PR "opening" comment. It's a fix for the scenario we discussed after standup yesterday and I mentioned in the doc. It allows you to do an auto paginated `sort({'_kmd.ect': 1}).limit(21000)`. You reminded me they can still do it by calling pull() with no auto pagination a few times (for when they want >10k results). But I already had it in there. And I feel it doesn't do harm, but I will remove it if it doesn't seem useful and/or is confusing.

src/core/datastore/sync/sync-manager.js
georgiwe: It isn't a pagination query, they are all created in one place. There was some back and forth with building/sanitizing of queries.

Its goal was to "sanitize" the user query and make it so that all internal pagination methods can use the queries being passed to them safely. It used to add the implicit sort, but looks like we're not going that route. Now its main purpose is adding the limit, when the user supplies it in the pullOptions. This is important for the local delete(), in order to avoid deleting everything.

I may rearrange things a bit today.

georgiwe:
Will open a new PR, with the changes from the extended spec.

PR #257, opened 2018-02-26
iOS Keychain conformance

heyzooi:
#### Description

JS SDK using the same params used by the Swift SDK

#### Changes

Using the same query params used by our Swift SDK

### DIFF
--- a/src/core/client.js
+++ b/src/core/client.js
@@ -170,7 +170,7 @@ export class Client {
    * Set the active user
    */
   setActiveUser(activeUser) {
-    if (activeUser != null) {
+    if (activeUser) {
       return this.activeUserStorage.set(`${this.appKey}.${ACTIVE_USER_KEY}`, activeUser);
     } else {
       return this.activeUserStorage.remove(`${this.appKey}.${ACTIVE_USER_KEY}`);
--- a/src/core/client.js
+++ b/src/core/client.js
@@ -170,11 +170,7 @@ export class Client {
    * Set the active user
    */
   setActiveUser(activeUser) {
-    if (activeUser) {
-      return this.activeUserStorage.set(`${this.appKey}.${ACTIVE_USER_KEY}`, activeUser);
-    } else {
-      return this.activeUserStorage.remove(`${this.appKey}.${ACTIVE_USER_KEY}`);
-    }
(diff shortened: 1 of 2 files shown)


heyzooi:
this PR is not required to fix the keychain issue, but conforms to our Swift SDK by using the same params.

georgiwe:
I don't fully understand keychain and iOS in general, but I'm thinking "will there be a backwards compatibility issue"? Users would have to log in again? Is there any necessary clean up of the old stuff in the keychain or something? Can't we "migrate" those?

thomasconner:
@heyzooi @georgiwe Georgi is right. These changes will require a user to login again. I think these changes make sense so that we use the same params as our Swift SDK but we need to think of how we will migrate existing active users. I will the missing backwards compatibility code.

heyzooi:
right, let me refactor that, that's really no need to keep the same keys, but the security changes are more important to be in place

src/core/client.js
```diff
@@ -158,7 +158,11 @@ export class Client {
    * Set the active user
    */
   setActiveUser(activeUser) {
-    return this.activeUserStorage.set(`${this.appKey}.${ACTIVE_USER_KEY}`, activeUser);
+    if (activeUser != null) {
```
heyzooi: question that I had was if should use `!=` or `!==` in this case

thomasconner: you can just do `if (activeUser) {`

heyzooi: done! thanks @thomasconner

georgiwe:
I see this is approved - shouldn't we fix the unit tests first?

tejasranade:
@thomasconner @heyzooi Who has the next steps on this PR? The unit tests are currently failing. We can't merge until they are fixed. Maybe this needs a JIRA?

thomasconner:
We need to look into why the unit tests are failing and make any changes that are needed. I don't know what changes are needed if any at the moment.

heyzooi:
@thomasconner should I merge this?

heyzooi:
@thomasconner should I rebase to see if the unit tests are not working?

heyzooi:
unit test are running again, @thomasconner i think you can now approve

PR #260, opened 2018-03-01
Auto pagination

georgiwe:
#### Description
MLIBZ-2045. Implement auto pagination. Unit tests are pending.

#### Changes
Delete unused file. Small change for eslint. Small improvements on jsdocs. Change the way defaults are set for Query class, so undefined values don't overwrite null values (which causes runtime errors).


src/core/datastore/cachestore.js
```diff
@@ -125,24 +128,21 @@ export class CacheStore extends NetworkStore {
...
-            };
-            return result;
-          });
-        return promise;
+      .then((pushResult) => {
+        result.push = pushResult;
+        return this.pull(query, options);
+      })
+      .then((pullResult) => {
+        result.pull = pullResult;
```
tejasranade: The signature of the method suggests the return parameters as an array (`<{push: [], pull: []}>`). This will return a number, if I understand it correctly.

src/core/datastore/sync/sync-manager.js
```diff
@@ -60,6 +66,12 @@ export class SyncManager {
     if (!isNonemptyString(collection)) {
       return Promise.reject(new KinveyError('Invalid or missing collection name'));
     }
+
+    // TODO: decide on default value of pagination setting
+    if (options && (options.autoPagination && !options.useDeltaFetch)) {
```
tejasranade: Using `useDeltaFetch` here seems incorrect. Whether to get the entire set of objects or a delta, should be decided based on the `useDeltaFetch` value and the `_QueryCache`. 

If I have run this query before (regardless of whether I ran it as "auto-paginated" or not) _and_ I have delta sync turned on, then make a delta request for me.

src/core/datastore/sync/sync-manager.js
```diff
@@ -347,4 +358,60 @@ export class SyncManager {
...
     }
   }
+
+  _getInternalPullQuery(userQuery = {}, totalCount) {
+    const { filter, sort, fields, skip } = userQuery;
+    const query = new Query({ filter, sort, fields, skip });
+    query.limit = totalCount;
+    const noSortProvided = !sort || isEmpty(sort);
+    const noSkipLimitProvided = !userQuery.skip && !userQuery.limit;
+    if (noSortProvided || noSkipLimitProvided) {
```
tejasranade: Shouldn't this be `&&`? The way it reads - if a user provided a sort, but not a skip/limit, we will replace their sort with ours.

src/core/datastore/cachestore.js
georgiwe: Yeah, that's from "the old branch", I'll update it.

src/core/datastore/sync/sync-manager.js
georgiwe: Well yeah, but delta set isn't implemented here yet. This will have to be done after delta set is ready.

georgiwe: Also, what's the decision on the setting - setting, no setting, default value?

thomasconner: Can you explain what you mean "setting, no setting, default value"?

georgiwe: There is a discussion in the second spec doc about whether auto pagination should be a setting or not. And if it is a setting, what should be its default value? There is no final decision yet. I've left it as a seeting, with a default value of `false`, as per the original spec doc.

src/core/datastore/sync/sync-manager.js
thomasconner: This doesn't make sense to me. I read the doc but I am still confused here.

georgiwe: We can discuss it at standup. It's the second bullet point of the section about sort.

src/core/datastore/cachestore.js
```diff
@@ -99,12 +99,15 @@ export class CacheStore extends NetworkStore {
    * Pull items for the data store from the network to your local cache. A promise will be
    * returned that will be resolved with the result of the pull or rejected with an error.
    *
+   * IMPORTANT: This method is not intended to be used to make concurrent requests.
```
georgiwe: This is about MLIBZ-2130. You may want to document this better, I think this is what we agreed upon in the comments of the ticket.

src/core/datastore/persisters/browser/websql-key-value-store-persister.js
```diff
@@ -1,153 +0,0 @@
-import { KinveyError, NotFoundError } from '../../errors';
```
georgiwe: This file is not used anywhere, it's an outdated file, left from all the rebasing during the redesign. In case anyone is wondering :)

src/core/platform-configs/configs.js
```diff
@@ -0,0 +1,19 @@
...
+    maxConcurrentPushRequests: 100,
+    maxConcurrentPullRequests: 32
+  },
+
+  [platformName.nativeScript]: {
+    maxConcurrentPullRequests: mobilePlatformsConcurrentPulls
+  },
+
+  [platformName.phoneGap]: {
+    maxConcurrentPullRequests: mobilePlatformsConcurrentPulls
```
georgiwe: I haven't tested on PhoneGap. I set it to the same as NativeScript, just in case. We can test it out before GA.

src/core/query.js
```diff
@@ -800,7 +801,7 @@ export class Query {
 
     // Remove fields
     if (Array.isArray(json.fields) && json.fields.length > 0) {
-      const fields = Array.concat([], json.fields, PROTECTED_FIELDS);
+      const fields = [].concat(json.fields, PROTECTED_FIELDS);
```
georgiwe: This will cause a conflict once #258 gets merged. I haven't merged 258 cause of Tejas' request for changes. I keep forgetting to bring this up.

test/integration/tests/utilities.js
```diff
@@ -59,6 +59,7 @@
       return networkStore.save(entity);
     }))
       .then(() => syncStore.pull())
+      .then(() => syncStore.find().toPromise())
```
georgiwe: @tsvetomir-nedyalkov please have a look at this and make sure it won't change the behaviour of the method. The result type of pull() has changed and it shouldn't be used like it is being used here. Yes, it is a breaking change, but it's been discussed and approved.

test/integration/tests/utilities.js
```diff
@@ -181,6 +182,11 @@
       .then(() => syncStore.clear());
   }
 
+  function cleanAndPopulateCollection(collectionName, entities) {
```
georgiwe: @tsvetomir-nedyalkov I noticed `cleanUpCollectionData()` and `saveEntities()` are almost always used together, so I added this helper method.

PR #264, opened 2018-03-09
Fix Mobile Identity Connect Popup Issues

thomasconner:
#### Description
The Angular, Angular2, and PhoneGap SDKs all share the same popup class. All the imports expected an exported class named `Popup` but instead it was named `PhoneGapPopup`. This would cause a MIC to not work on environments using these SDKs.

#### Changes
- Rename `PhoneGapPopup` to `Popup`.
- Instantiante the `Popup` class when requesting a code for Mobile Identity Connect login.

### DIFF
--- a/src/phonegap/popup.js
+++ b/src/phonegap/popup.js
@@ -93,7 +93,7 @@ export class Popup extends HTML5Popup {
   }
 
   static open(url) {
-    const popup = new PhoneGapPopup();
+    const popup = new Popup();
     return popup.open(url);
   }
 }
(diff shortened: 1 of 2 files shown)


src/phonegap/popup.js
```diff
@@ -1,7 +1,7 @@
-import { Popup } from '../html5/popup';
+import { Popup as HTML5Popup } from '../html5/popup';
 import { Device } from './device';
 
-export class PhoneGapPopup extends Popup {
+export class Popup extends HTML5Popup {
```
tejasranade: There is another place in this file that has `PhoneGapPopup`. Look for the `static open(url)` method at the bottom. That needs to change too

thomasconner: Yep good catch.

PR #265, opened 2018-03-12
Beta release bugfixes

georgiwe:
#### Description
Fixes for MLIBZ-2398 and MLIBZ-2405. One was a typo, remaining from before the change in the name of the constant. The other was a suboptimal default value we used in the shorthand for executing an HTTP request. @thomasconner please confirm this is the way it should be used. It's overwritable, but I believe this should be the (aptly named) default always, no?

#### Changes
Fix name of "enum" for platform names. Use shallow copies of entities when using inmemory persistance. Change the default auth parameter of network request config.

### DIFF
--- a/src/core/datastore/persisters/memory-key-value-persister.js
+++ b/src/core/datastore/persisters/memory-key-value-persister.js
@@ -10,11 +10,12 @@ export class MemoryKeyValuePersister extends KeyValuePersister {
   }
 
   _readFromPersistance(key) {
-    return Promise.resolve(_storage[key]);
+    const result = _storage[key] || [];
+    return Promise.resolve(result.slice(0));
   }
 
   _writeToPersistance(key, array) {
-    _storage[key] = array;
+    _storage[key] = array.slice(0);
     return Promise.resolve(true);
   }
 
--- a/src/core/request/network.js
+++ b/src/core/request/network.js
@@ -165,7 +165,7 @@ export class KinveyRequest extends NetworkRequest {
   static execute(options, client, dataOnly = true) {
     const o = assign({
       method: RequestMethod.GET,
-      authType: AuthType.Session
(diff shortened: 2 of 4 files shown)


src/core/datastore/persisters/memory-key-value-persister.js
```diff
@@ -10,11 +10,12 @@ export class MemoryKeyValuePersister extends KeyValuePersister {
...
 
   _readFromPersistance(key) {
-    return Promise.resolve(_storage[key]);
+    const result = _storage[key] || [];
+    return Promise.resolve(result.slice(0));
   }
 
   _writeToPersistance(key, array) {
-    _storage[key] = array;
+    _storage[key] = array.slice(0);
```
thomasconner: Why was this change made?

src/core/request/network.js
```diff
@@ -165,7 +165,7 @@ export class KinveyRequest extends NetworkRequest {
   static execute(options, client, dataOnly = true) {
     const o = assign({
       method: RequestMethod.GET,
-      authType: AuthType.Session
+      authType: AuthType.Default
```
thomasconner: This is the correct way to set the default `AuthType`.

src/core/datastore/persisters/memory-key-value-persister.js
georgiwe: To avoid changing objects by reference. This was the reason one of the unit tests, which are currently skipped, was failing.

It was being called twice with the same array, except the second time it had another element in it. So at the time when the two calls to the spy were being compared, the array had 3 elements, instead of two. This is essentially how other persisters work too - when you read from localStorage, of instance, you always get a new object. Whereas here with inmemory persister, you got the same array each time you read. So it was being modified and everyone who has ever read it and kept it, got a modified array.

PR #267, opened 2018-03-12
Fix Delta Set Issues

thomasconner:
#### Changes
- Properly generate `_id` for query cache doc.
- Handle the updating of the `_QueryCache` whenever a `pull()` is made.
- Update tests.


src/core/datastore/repositories/network-repository.js
```diff
@@ -21,9 +23,22 @@ import { buildCollectionUrl } from './utils';
  */
 
 export class NetworkRepository extends Repository {
-  read(collection, query, options = {}) {
+  read(collection, query, options = {}, responseCallback) {
```
georgiwe: It was my first thought to have the HTTP data when working with `NetworkRepository`, but now I don't think this is the best approach. We should leave the `NetworkRepository` be a repository like any other. If it isn't the tool for the job, we should use a different abstraction.

We should not need any HTTP-related details when we work with a repository. Users of the repository only care about working with entities/collections. If we need HTTP-specific data, we should use something else. What this is, is to be determined by the use case and we should think in this direction.

PR #268, opened 2018-03-13
Add scope parameter to mic for v3.

tejasranade:
#### Description
After discussion with @andrewmaillet , we realized that v3 requires a `scope` parameter in the `/oauth/auth` call. This PR adds it.


src/core/identity/mic.js
```diff
@@ -204,7 +205,8 @@ export class MobileIdentityConnect extends Identity {
           username: options.username,
           password: options.password
         },
-        followRedirect: false
+        followRedirect: false,
+        scope: 'openid'
```
thomasconner: This should be part of the body parameters.

PR #270, opened 2018-03-14
MLIBZ-2316: Server Side Delta Set

thomasconner:
#### Description
Add support for server side delta set.

#### Changes
- Rename `deltaFetch.js` to `deltaSet.js` and move it to under `src/core/datastore`. This no longer extends the `Request` class.
- `deltaSet.js` add support for sending delta set requests and clearing the query cache.
- Rename `option.useDeltaFetch` to `option.useDeltaSet` on datastore instances. @tejasranade says this does not need to be backwards compatible.

#### Test
- Added tests to `cachestore.spec.js` for sending a delta set request with and without a tagged collection.


src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+            url = format({
+              protocol: client.apiProtocol,
+              host: client.apiHost,
+              pathname: `/appdata/${client.appKey}/${collectionName}/_deltaset`,
+              query: { since: deltaSetQueryDoc.lastRequest }
+            });
+          }
+
+          const request = new KinveyRequest({
+            authType: AuthType.Session,
```
georgiwe: This shouldn't be allowed when using the SDK with master secret auth?

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+            .then((response) => {
+              deltaSetQueryDoc.lastRequest = response.headers.get('X-Kinvey-Request-Start');
+              return offlineRepo.update(QUERY_CACHE_COLLECTION_NAME, deltaSetQueryDoc)
+                .then(() => response.data);
+            });
+        })
+        .then((data) => {
+          const { deleted } = data;
+
+          if (isArray(deleted) && deleted.length > 0) {
```
georgiwe: This check is to check if `data` is the result of a regular GET or a _deltaset request. I think writing things separately and knowing which request was made above is a better design and improves readability. This ties in with the comment about the size of this function - if it were broken up in smaller functions by responsibility, this wouldn't happen.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+            return offlineRepo.delete(collectionNameWithTag, deleteQuery)
+              .then(() => data);
+          }
+
+          return data;
+        })
+        .then((data) => {
+          let { changed } = data;
+
+          if (!changed) {
```
georgiwe: Same as the one about line 73.

src/core/request/deltafetch.spec.js
```diff
@@ -1,316 +0,0 @@
-import nock from 'nock';
```
georgiwe: So now we only have one test for delta set? Will there be more added?

src/core/datastore/repositories/network-repository.js
```diff
@@ -22,8 +23,12 @@ import { buildCollectionUrl } from './utils';
 
 export class NetworkRepository extends Repository {
   read(collection, query, options = {}) {
+    if (options.useDeltaSet) {
+      return deltaSet(collection, query, options);
```
georgiwe: I believe this returns only the modified entities. I don't think this is the spec for delta set in CacheStore.find(). Again, I don't have an app to test it, just judging from the code.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+            followRedirect: options.followRedirect,
+            cache: options.cache,
+            properties: options.properties,
+            skipBL: options.skipBL,
+            trace: options.trace,
+            client
+          });
+          return request.execute()
+            .then((response) => {
+              deltaSetQueryDoc.lastRequest = response.headers.get('X-Kinvey-Request-Start');
```
georgiwe: Maybe a constant for the header name?

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
+import isArray from 'lodash/isArray';
+import { format } from 'url';
+import { Query } from '../query';
+import { KinveyRequest, RequestMethod, AuthType } from '../request';
+import { Client } from '../client';
+import { repositoryProvider } from './repositories';
+import { stripTagFromCollectionName, generateEntityId } from './utils';
+
+const QUERY_CACHE_COLLECTION_NAME = '_QueryCache';
```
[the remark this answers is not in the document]
tejasranade: I feel differently about the naming. Calling it `_QueryCache` in the design was a deliberate decision. I am ok with alternatives that describe what it is - it is a log of queries and corresponding timestamps indicating the last time they were run.  `_QueryHistory`, `_QueryLog`, or equivalent is fine by me.

I want to avoid naming it based on the "delta sync" feature. We could use this table later for other purposes.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+
+export function deltaSet(collectionNameWithTag, query, options) {
```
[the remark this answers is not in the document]
tejasranade: +1 on both comments. I left Victor a very similar comment on the iOS deltaset implementation.
https://github.com/Kinvey/swift-sdk/pull/261#discussion_r168624898

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+  const collectionName = stripTagFromCollectionName(collectionNameWithTag);
+  const serializedQuery = query ? JSON.stringify(query.toQueryString()) : '';
+
+  return repositoryProvider.getOfflineRepository()
+    .then((offlineRepo) => {
+      const queryCacheQuery = new Query()
+        .equalTo('collectionName', collectionName)
```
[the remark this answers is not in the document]
tejasranade: I agree wholeheartedly with the middle paragraph. 
>  Only places which make a network request should know about a tag, in order to remove it when making the request.

This should be the goal.

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
thomasconner: Yep this makes sense. I will break up the function into smaller functions.

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
georgiwe: I hadn't thought of other uses for it. I guess there could be some in the future, I don't know. But "query log" sounds pretty good to me.

I think when the word "cache" is used, it's usually something that has a TTL and what's more important about it is that it is not essential to the work of the system - anything that uses cache, should function without cache, just slower, probably. So I think it would be nice to move away from naming the offline stores "caches" in general. But that's a separate topic :)

src/core/datastore/cachestore.spec.js
```diff
@@ -190,6 +190,80 @@ describe('CacheStore', () => {
...
+        .reply(200, [entity1, entity2], {
+          'X-Kinvey-Request-Start': lastRequestDate.toISOString()
+        });
+
+      store.pull()
+        .then(() => {
+          nock(client.apiHostname)
+            .get(`/appdata/${store.client.appKey}/${collection}/_deltaset`)
+            .query({ since: lastRequestDate.toISOString() })
+            .reply(200, { changed: [entity2], deleted: [{ _id: entity1._id }] }, {
```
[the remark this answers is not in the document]
thomasconner: I added a change to `entity2` and check if that change is made.

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
thomasconner: I broke up `deltaSet()` into smaller functions.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+import { KinveyRequest, RequestMethod, AuthType } from '../request';
+import { Client } from '../client';
+import { repositoryProvider } from './repositories';
+import { stripTagFromCollectionName, generateEntityId } from './utils';
+
+const QUERY_CACHE_COLLECTION_NAME = '_QueryCache';
+
+export function deltaSet(collectionNameWithTag, query, options) {
+  const collectionName = stripTagFromCollectionName(collectionNameWithTag);
+  const serializedQuery = query ? JSON.stringify(query.toQueryString()) : '';
```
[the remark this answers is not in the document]
thomasconner: An error is now thrown if a query with `skip` and `limit` modifiers is used.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,99 @@
...
+            collectionName: collectionName,
+            query: serializedQuery
+          };
+        })
+        .then((deltaSetQueryDoc) => {
+          const client = Client.sharedInstance();
+          let url = format({
+            protocol: client.apiProtocol,
+            host: client.apiHost,
+            pathname: `/appdata/${client.appKey}/${collectionName}`
```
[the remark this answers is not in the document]
thomasconner: Yes it was. I am using it now.

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
thomasconner: I changed it to use the full collection name which includes the tag.

src/core/datastore/deltaset.js
thomasconner: I changed it to use `AuthType.Default`.

src/core/datastore/deltaset.js
thomasconner: I moved the logic into the correct smaller function.

src/core/datastore/deltaset.js
thomasconner: Same as above.

src/core/request/deltafetch.spec.js
thomasconner: I added 2 more for `pull()`. I will probably add some for `deltaSet.js`

src/core/datastore/repositories/network-repository.js
thomasconner: You were right. It now returns what matches the provided query in cache after the requests are complete.

src/core/datastore/deltaset.js
thomasconner: I added a const to use for the header name.

src/core/datastore/sync/sync-manager.js
```diff
@@ -291,7 +291,6 @@ export class SyncManager {
   }
 
   _fetchItemsFromServer(collection, query, options) {
```
georgiwe: Wouldn't this result in redundancies? By "this" I mean

```javascript
return this._fetchItemsFromServer(collection, query, options)
      .then(entities => this._replaceOfflineEntities(collection, query, entities))
      .then(replacedEntities => replacedEntities.length);
```

in `SyncManager` (it's not in the diff). This is the old code and it seems like the updating and deleting (the `_replaceOfflineEntities()` part) is now being done inside the deltafetch "module". Same goes for when doing a CacheStore.read(). The old `DeltaFetchRequest` only did a read, the new one does writes too. I guess something else has to change to reflect this too. These redundant deletes and updates would result in worse performance - they are very expensive since we're working in memory.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+          return {
+            _id: generateEntityId(),
+            collectionName: collectionNameWithTag,
+            query: serializedQuery
+          };
+        });
+    });
+}
+
+function updateDeltaSetQuery(deltaSetQuery, response) {
```
georgiwe: This function doesn't seem to need the response. It only needs the header value.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+      return updateDeltaSetQuery(deltaSetQuery, response).then(() => response.data);
+    })
+    .then((data) => {
+      return deleteDocs(collectionNameWithTag, data.deleted)
+        .then(() => updateDocs(collectionNameWithTag, data.changed))
+        .then(() => data);
+    });
+}
+
+function makeRequest(collectionNameWithTag, deltaSetQuery, query, options) {
```
georgiwe: This function can very easily be split into two - it does two (and a half) things - performs a delta set request, or a regular get request (and the "half" is managing how local store should be updated). It's clearly visible where one, self-sufficient part ends (the first `return`). This should be a separate function.

The name "makeRequest" even suggests things should be different - if you can't think of a descriptive name for a function, it means it's not clear what it does. Which means it should change.

I think the way to solve this is split `makeRequest` into all of its small operations and then "merge" `makeRequest` with `deltaSet`. The resulting function would delegate all operations to other functions and handle common context. This is readable and each separate function would be cohesive.

After "a request" is made, both "paths" (which should be functions) call `updateDeltaSetQuery()`. This would be solved by the above paragraph too.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+      return updateDeltaSetQuery(deltaSetQuery, response).then(() => response.data);
+    })
+    .then((data) => {
+      return updateDocs(collectionNameWithTag, data);
+    });
+}
+
+export function deltaSet(collectionNameWithTag, query, options) {
+  return getDeltaSetQuery(collectionNameWithTag, query)
+    .then((deltaSetQuery) => makeRequest(collectionNameWithTag, deltaSetQuery, query, options))
```
georgiwe: There was talk about renaming this "query cache" table to "query log" - is this rejected, @tejasranade ? This would make the name of the variable `deltaSetQuery` (and related functions) something like `queryLogEntry`, which is more descriptive in my opinion. Calling it "query" right now is incorrect - it isn't a query, a serialized query is part of it. I know you'll say this is all variable names, but it isn't - it's terms which will be used and they should be well defined.

src/core/datastore/sync/sync-manager.js
georgiwe: Actually, it's just the updating that's always done. The deleting is done only for when doing a delta set request.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+function getDeltaSetQuery(collectionNameWithTag, query) {
+  const serializedQuery = query ? query.toString() : '';
+
+  return repositoryProvider.getOfflineRepository()
+    .then((offlineRepo) => {
+      const queryCacheQuery = new Query()
+        .equalTo('collectionName', collectionNameWithTag)
+        .and()
+        .equalTo('query', serializedQuery);
+      return offlineRepo.read(queryCacheCollectionName, queryCacheQuery)
```
georgiwe: I now remember we went a different route with the `kinvey_sync` collection - it is being scoped like any other collection. Here with the "query cache" collection, we are only scoping the collection name of its "rows". Should we may be unify our approach?

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+
+function updateDeltaSetQuery(deltaSetQuery, response) {
+  return repositoryProvider.getOfflineRepository()
+    .then((offlineRepo) => {
+      deltaSetQuery.lastRequest = response.headers.get(xKivneyRequestStartHeader);
+      return offlineRepo.update(queryCacheCollectionName, deltaSetQuery);
+    });
+}
+
+function readDocs(collectionNameWithTag, query) {
```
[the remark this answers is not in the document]
tejasranade: "entities" is the term we generally use.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+  return repositoryProvider.getOfflineRepository()
+    .then((offlineRepo) => {
+      return offlineRepo.read(collectionNameWithTag, query);
+    });
+}
+
+function deleteDocs(collectionNameWithTag, deleted = []) {
```
tejasranade: This method signature is confusing. Is the purpose of the method to `delete` the entities that are passed in as an array parameter? If so, it's clearer to name that variable something like `entitiesToDelete` rather than `deleted`.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+      }
+
+      return changed;
+    });
+}
+
+function makeDeltaSetRequest(collectionNameWithTag, deltaSetQuery, query, options) {
+  const collectionName = stripTagFromCollectionName(collectionNameWithTag);
+  const client = Client.sharedInstance();
+  const request = new KinveyRequest({
```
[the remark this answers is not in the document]
tejasranade: @georgiwe I don't completely understand your comment. What `KinveyRequest.execute()` call are you referring to? Is there an example.

I agree in principle with your point, though. It seems like a bunch of these options (`skipBL`, `timeout` etc) would apply to all data requests, so some way to avoid listing them out one-by-one will be useful.

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+    .then((data) => {
+      return deleteDocs(collectionNameWithTag, data.deleted)
+        .then(() => updateDocs(collectionNameWithTag, data.changed))
+        .then(() => data);
+    });
+}
+
+function makeRequest(collectionNameWithTag, deltaSetQuery, query, options) {
+  if (deltaSetQuery && deltaSetQuery.lastRequest) {
+    if (query && (query.skip != null || query.limit != null)) {
```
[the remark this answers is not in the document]
tejasranade: @georgiwe re: your second point - isn't the idea that `limit` and `skip` together with a deltaset request is not supported? If I'm reading this code snippet correctly, making a `deltaSetQuery` with `limit` and `skip` is not supported. I think that's reasonable. Thoughts?

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+      throw new KinveyError('You cannot use the skip and limit modifiers on the query when performing a delta set request.');
```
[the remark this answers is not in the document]
tejasranade: Aren't they equivalent, in how we're using them? I wasn't quite sure of the difference, so I looked up this post - https://stackoverflow.com/questions/33445415/javascript-promises-reject-vs-throw

Now I think our use of `throw (err)` seems ok. WDYT?

src/core/datastore/deltaset.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    return makeDeltaSetRequest(collectionNameWithTag, deltaSetQuery, query, options);
```
tejasranade: @georgiwe made a point about the `makeRequest` naming earlier in the review, and I find it relevant here. If the same class has `makeRequest` and `makeDeltaSetRequest` functions, it's not clear what each is doing.

src/core/datastore/deltaset.js
tejasranade: The design refers to the additional table as `_QueryCache`. Since there was some concern about whether it was a "cache", I floated the idea of `_QueryLog` instead. TBH, I suggest we go with `_QueryCache` because that's what we're using in other platforms.

In that context, we could use the names - `getCachedQuery` and `cachedQuery` here.

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
georgiwe: I mean like [here](https://github.com/Kinvey/js-sdk/blob/master/src/core/live/collection/live-collection-manager.js#L70). It basically takes care of all these defaults which are used over 90% of the time, extracting the data, etc. Just getting rid of boilerplate.

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
georgiwe: If an exception is thrown somewhere in a promise chain, it's essentially the same. Like so:

```javascript
function someFunc() {
  throw new Error('aaa');
}

Promise.resolve()
  .then(() => {
    someFunc();
  })
  .catch(err => {
    console.log('caught in promise.catch()', err.message); // the thrown error
  });
```

But, if this function is called outside of a promise chain like so:

```javascript
try {
  someFunc() // throws a synchronous exception
    .catch(err => {
      console.log('err', err.message); // doesn't fire
    });
} catch (err) {
  console.log('caught in catch', err.message);
}
```

Currently, this particular function is only called in a promise, but it's not robust - if moved outside, things might break unexpectedly.

Another side of the reason is that a promise represents "a value in time". And the reason thrown exceptions are "caught" by promises is precisely this - they shouldn't be synchronous.

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
georgiwe: Sure, and I didn't think of this at the time, but yesterday I kinda changed my mind. And this may affect the way we treat skip and limit, if we want to keep some consistency.

It seems important to keep `.find()`s fully functional. In all store types, with all settings. If not, user would need to write their app in a less convenient way and keep track of what works when. If we start throwing an error in CacheStore.find() when skip and limit is passed, and deltaset is enabled, as a developer, it would be difficult to write my app. I may want to make some skip and limit requests - should I disable delta set for those requests only?

One way to keep them, is not send them as part of the delta set request (ignore them), but apply them to the request to the offline store, [here](https://github.com/Kinvey/js-sdk/blob/MLIBZ-2316_server_side_delta_set/src/core/datastore/deltaset.js#L152). This way we "update the local store", and then read whatever the user is trying to read, from a "synced" place.

This approach may introduce some confusion as it's different from `pull()`, but I think it isn't too bad.

Does this make sense, or am I missing something?

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
georgiwe: Oh! Totally forgot about the link you sent, sorry. Yeah, that use case is even trickier - throwing on the next processor tick. It's just a matter of not expecting a synchronous exception, when working with promises. Reading the function doesn't necessarily tell you "I should only use this in promise chains". We should do what essentially the example in the answer you linked does - wrap the exception in a promise.

src/core/datastore/deltaset.js
georgiwe: Alright, `_QueryCache` it is. Then all that's left is renaming `deltaSetQuery` to something like "queryCacheEntry", cause `cachedQuery` sounds like what would be in `queryCacheEntry.query`. So just one more term to "coin" - what is the entry in the query cache table :)

georgiwe: Okay, `_QueryCache` it is. And I think `getCachedQuery ` and `cachedQuery ` work in this context.

Just one more term to define - what is one entry in that table called? Cause it shouldn't be a "query" (and that's kinda why I didn't want the table to be a cache for queries).

src/core/datastore/sync/sync-manager.js
thomasconner: Should something be changed related to this?

src/core/datastore/deltaset.js
thomasconner: I don't understand what you mean. Can you explain more?

georgiwe: I mean that when you have a data store with no tag, it uses a collection `kinvey_sync` for the sync queue. If you use a data store with a tag, for the same collection, it uses a collection called `kinvey_sync.<tag-name>` for the sync queue.

The same two cases, regarding the `_QueryCache` collection, would use the same collection. Just the values for the `collection` "column" would be different. The same could have been applied to the `kinvey_sync` collection, but we went a different route. I'm saying we should decide which is better, and a unified behaviour might be a consideration in this decision.

src/core/datastore/deltaset.js
```diff
@@ -137,23 +117,40 @@ function makeRequest(collectionNameWithTag, deltaSetQuery, query, options) {
...
+        }
+
+        promise = makeDeltaSetRequest(collectionName, cachedQuery.lastRequest, query, options);
+      } else {
+        promise = makeRegularGETRequest(collectionName, query, options);
+      }
+
+      return promise
+        .then((response) => {
+          const requestStartDate = response.headers.get(xKivneyRequestStartHeader);
```
tejasranade: Just noticed a typo here while randomly browsing code - `xKivneyRequestStartHeader` :)

src/core/datastore/sync/sync-manager.js
thomasconner: I made a change related to this.

src/core/datastore/deltaset.js
georgiwe: I think we don't need the "x" in there too, while we're on the topic :)

src/core/datastore/sync/sync-manager.js
thomasconner: Should we be adding this same check to `CacheStore.find()`?

thomasconner:
I think all requested changes have been made. Is there anything else missing that is preventing this PR from being approved?

src/core/datastore/deltaset.js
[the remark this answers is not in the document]
thomasconner: This makes sense. I think your suggestion also makes sense. @tejasranade thoughts?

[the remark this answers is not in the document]
georgiwe: Yes, this is what I'm suggesting. I don't think it conflicts. I think skip and limit help you target a specific subset of data (just like you said), usually when paging things, but not always. Not just from the server, but from any data source. Cause if you had to pull everything, just to take the last 3 entities, things don't make sense, regardless of the data source.

I guess turning off deltaset would make the API easier to work with, but I think what I suggested is better - we get the best of both worlds. We get performance, we get the expected result and an easy to use API. What do you dislike about it - you feel it's semantically wrong?

Ultimately, I'm fine with whatever you decide - the most important part is to have an easy to use API, which works correctly and as expected.

thomasconner:
@Tejas @Martin-Hristov Please start reviewing this PR as I think I have finally fixed the remaining issues.

thomasconner:
@tejas I fixed the skipped unit tests.

thomasconner:
Can I have this PR looked over again? There have been several bug fixes made over the past few days related to testing.

tejasranade:
@thomasconner this branch needs a merge from master. Once you do that, we can do a final review before merging.

thomasconner:
Master has been merged.

PR #271, opened 2018-03-15
MLIBZ-2366: Fix Mobile Identity Connect on NativeScript

thomasconner:


src/nativescript/user/user.android.js
```diff
@@ -0,0 +1,36 @@
+import { User as CoreUser } from '../../core/user';
+
+export class User extends CoreUser {
```
tejasranade: Looks suspicious. Intentionally empty?

src/nativescript/user/user.ios.js
```diff
@@ -0,0 +1,24 @@
+import { User as CoreUser } from '../../core/user';
+
+export class User extends CoreUser {
+  static handleMICRedirectURL(redirectUri, micRedirectURL) {
```
tejasranade: Who calls this method? I don't see a reference to this method name in the current master. I don't see any invocations in this PR either. What am I missing?

src/nativescript/user/user.android.js
thomasconner: This is no longer needed. I will remove it.

src/nativescript/user/user.ios.js
thomasconner: This is no longer needed. I will remove it.

PR #276, opened 2018-03-23
MLIBZ-2131 Remove implicit push() calls

georgiwe:
#### Description
Remove the implicit push() calls in DataStore.pull(), CacheStore.find(), CacheStore.findById(), CacheStore.count(), CacheStore.group(). This removes potentially unexpected behaviour and improves the number of local reads from 2 to 4, to always 2.

#### Changes
Remove the count-push-count flow from CacheStore read-based ops. They do count and return an error if there are entities pending push. Otherwise proceed with their regular flow.

Put the check for entities pending push after the return of the offline result - it isn't being impeded by the entities pending sync.

Update tests accordingly.

Change the error message - input welcome.

### DIFF
--- a/src/core/datastore/utils/utils.js
+++ b/src/core/datastore/utils/utils.js
@@ -6,8 +6,13 @@ import { isNonemptyString } from '../../utils';
 export const dataStoreTagSeparator = '.';
 
 export function getEntitiesPendingPushError(entityCount, prefix) {
-  const countMsg = `There are ${entityCount} entities, matching this query or id, pending push to the backend.`;
-  const errMsg = `Unable to ${prefix} on the backend, since the result might overwrite your local changes. ${countMsg}`;
+  let countMsg = `There are ${entityCount} entities, matching the provided query, pending push to the backend.`;
+
+  if (entityCount === 1) {
+    countMsg = `There is ${entityCount} entity, matching the provided query or id, pending push to the backend.`;
+  }
+
+  const errMsg = `Unable to ${prefix} on the backend. The result will overwrite your local changes. ${countMsg}`;
   return new KinveyError(errMsg);
 }
 
--- a/src/core/datastore/utils/utils.js
+++ b/src/core/datastore/utils/utils.js
@@ -1,8 +1,15 @@
 import _isEmpty from 'lodash/isEmpty';
 
+import { KinveyError } from '../../errors';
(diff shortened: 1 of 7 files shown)


src/core/datastore/utils/utils.js
```diff
@@ -1,9 +1,16 @@
 import _isEmpty from 'lodash/isEmpty';
 
+import { KinveyError } from '../../errors';
 import { isNonemptyString } from '../../utils';
 
 export const dataStoreTagSeparator = '.';
 
+export function getEntitiesPendingPushError(entityCount, prefix) {
+  const countMsg = `There are ${entityCount} entities, matching this query or id, pending push to the backend.`;
```
thomasconner: Lets use an `if` statement to make the count message more grammatically correct.

```javascript
let countMsg = `There are ${entityCount} entities, matching the provided query, pending push to the backend.`;

if (entityCount === 1) {
  countMsg = `There is ${entitiyCount} entity, matching the provided query or id, pending push to the backend.`;
}
```

src/core/datastore/utils/utils.js
```diff
@@ -1,9 +1,16 @@
...
+  const errMsg = `Unable to ${prefix} on the backend, since the result might overwrite your local changes. ${countMsg}`;
```
thomasconner: I would suggest to break up the error message into two sentences.

```javascript
const errMsg = `Unable to ${prefix} on the backend. The result might overwrite your local changes.`;
```

tejasranade: I'm a little confused. When does `errMsg` get used? 
I think the revised version from Thomas is an improvement, but I'd like to avoid the phrase "might overwrite" if possible. Can we say something more concrete?

georgiwe: `errMsg` is the entire message - it's always used. Just there were a few, slightly different, error messages (thus "prefix" parameter).

The method exists just to build the error message. As we discussed, there will someday be a general error factory, but until then, this will have to suffice.

I guess "the returned entities would overwrite your local entities" is more concrete. How about it?

thomasconner: `The result will overwrite your local changes.` @tejasranade thoughts?

tejasranade: Cool, I think `will` or `can` sounds more concrete. @thomasconner let's go with yours.

PR #278, opened 2018-03-26
MLIBZ-2332: Add support for Instance ID

thomasconner:
#### Description
Allow a user to provide an`instanceId` config option to `init()` that will setup up the Kinvey API and MIC urls correctly.

```javascript
const client = init({
  appKey: '<appKey>',
  appSecret: '<appSecret>',
  instanceId: '<my-subdomain>'
});
client.apiHostname // https://<my-subdomain>-baas.kinvey.com
client.micHostname // https://<my-subdomain>-auth.kinvey.com
```

#### Changes
- Support config option `instanceId` in `client.js` constructor.

#### Tests
- Added a test to `client.spec.ts` that checks if an error is thrown if `instanceId` is anything but a string
- Added a test to `client.spec.ts` that checks if an `instanceId` is provided that the `apiHostname` and `micHostname` are correct.

### DIFF
--- a/src/core/client.js
+++ b/src/core/client.js
@@ -54,8 +54,8 @@ export class Client {
    *
    * @param {Object}    options                                            Options
    * @param {string}    [options.instanceId='<my-subdomain>']              Custom subdomain for Kinvey API and MIC requests.
-   * @param {string}    [options.apiHostname='https://baas.kinvey.com']    Deprecated: Host name used for Kinvey API requests
-   * @param {string}    [options.micHostname='https://auth.kinvey.com']    Deprecated: Host name used for Kinvey MIC requests
+   * @param {string}    [options.apiHostname='https://baas.kinvey.com']    Deprecated: Use the instanceID property instead. Host name used for Kinvey API requests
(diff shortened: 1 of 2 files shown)


src/core/client.js
```diff
@@ -70,12 +71,30 @@ export class Client {
...
+      const { instanceId } = config;
+
+      if (!isString(instanceId)) {
+        throw new KinveyError('Instance ID must be a string.');
+      }
+
+      apiHostname = `https://${instanceId}-baas.kinvey.com`;
+      micHostname = `https://${instanceId}-auth.kinvey.com`;
+    } else {
+      if (isString(config.apiHostname)) {
```
tejasranade: As part of this ticket, the `micHostname` and `apiHostname` config options should be marked deprecated in the API reference. Also, the `instanceId` option should be documented in the API reference.

thomasconner: I already did this. Take a look at the documentation comments above the function.

tejasranade: My mistake, I see it now. Can you make the following change? 
"Deprecated: Host name used for Kinvey API requests": add the following - "Use the `instanceID` property instead"

LGTM overall.

PR #281, opened 2018-03-28
Add default entitlements plist file to nativescript sdk

tejasranade:
#### Description
Refer to https://github.com/Kinvey/nativescript-sdk/issues/31 for details on the issue and and proposed fix.

#### Changes
This PR adds a default `app.entitlements` file to the nativescript-sdk.

#### Tests
Tested manually with a MIC sample (https://github.com/tejasranade/EnterpriseSSO)

### DIFF
--- a/packages/kinvey-nativescript-sdk/platforms/ios/app.entitlements
+++ b/packages/kinvey-nativescript-sdk/platforms/ios/app.entitlements
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>keychain-access-groups</key>
+    <array>
+        <string>$(AppIdentifierPrefix)$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+    </array>
+</dict>
+</plist>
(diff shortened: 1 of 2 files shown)


packages/kinvey-nativescript-sdk/platforms/ios/app.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>keychain-access-groups</key>
+    <array>
+        <string>$(AppIdentifierPrefix)$(PRODUCT_BUNDLE_IDENTIFIER)</string>
```
thomasconner: How did you determine this should be the value here?

tejasranade: Good question. I believe I got this as the recommended default. It's basically a way to create a group identifier without having to hardcode any part of it. Here's some info on it - https://stackoverflow.com/a/39437477
And this is a tutorial on what the value means - https://evgenii.com/blog/sharing-keychain-in-ios/
I'll try to find the official docs about this, but we've been using this same value in all our sample apps and templates.

thomasconner: Ok thanks for the explanation.

PR #282, opened 2018-04-06
MLIBZ-2429: Don't include push by default for NativeScript SDK

thomasconner:
#### Description
Removes the import of the `Push` module in the main bundle. A separate push bundle is created that can be imported when push support is needed.

```js
import KinveyPush from 'kinvey-nativescript-sdk/push';
```

#### Changes
- Remove `export { Push } from './push'` from main bundle file.
- Create separate push bundle that exports push only.
- Update webpack configs for iOS and Android to create push bundle.

#### Tests
- Tested manually on iOS and Android


thomasconner:
This fixes https://github.com/NativeScript/kinvey-integration-issues/issues/4.

thomasconner:
I need to update the related DevCenter docs. I will link to this PR when I create the PR for the DevCenter changes.

georgiwe:
I'm not entirely clear why the issue happens, but I have a couple of thoughts here.

- Isn't this a breaking change? Is there no other way to fix this? I think it's worth trying to "trick" webpack by using a variable, instead of a string when calling `require('nativescript-push-plugin')`. Dunno if it would work, but it may be worth a try, to avoid a breaking change. The call to `require` can also be made lazy, to avoid immediate initialization, if that's also a concern.

- I think this would about double the size of the SDK. If we're going the route of a second bundle, perhaps we should look into optimizing the webpack config, to share code, common to both bundles. Some of the big problems of NS are large app size and slow build time - this exacerbates both of these.

georgiwe:
After some basic testing, using a variable for the require function or the name of the dependency seems to make webpack ignore it in the bundle, but when the app does use push functionality, we have no way to make it bundle it. Since it's the NS CLI that makes the bundling, we have no control. So we may have to make this breaking change indeed.

PR #284, opened 2018-04-10
MLIBZ-2307: NativeScript Push Unregister Bug

thomasconner:
#### Description
Unregistering from push will fail on Android because the options used are incorrect.


thomasconner:
Yes I think we need to.

PR #289, opened 2018-05-01
MLIBZ-2446: Add support for kinveyfile_ttl and kinveyfile_tls query parameter for Datastore GET requests

thomasconner:
#### Description
Files can be referenced in entities stored on a datastore collection. For more information please read [Files REST Guide](https://devcenter.kinvey.com/rest/guides/files#Linkingfilestoentitiesincollections). These file references are resolved when a `GET` request is made. The `GET` request can optionally set `kinveyfile_ttl` and/or `kinveyfile_tls` to change the `_downloadUrl` from `http` to `https` and limit how long the `_downloadUrl` is valid.

#### Changes
- Add support for `kinveyfile_ttl` and `kinveyfile_tls` query parameter `GET` requests made with the `NetworkRepository`
- Add unit tests to `networkstore.spec.js`, `cachestore.spec.js`, and `sync.spec.js` that check `kinveyfile_ttl` and `kinveyfile_tls` query parameters are properly set for `GET` requests.


thomasconner:
`kinveyfile_ttl` is used when you are loading an entity that has a file reference with a data store. `_expiresAt` is the value returned by the server that informs the user when the `_downloadUrl` is to expire. `ttl_in_seconds` is the query parameter used when downloading files using the `Kinvey.Files` API.

`tls` is the query parameter used when downloading files using the `Kinvey.Files` API. `kinveyfile_tls` is used when you are loading an entity that has a file reference with a data store.

Hopefully that makes sense. 😄

PR #290, opened 2018-05-02
Changed Facebook password in order to fix MIC Login tests

tsvetomir-nedyalkov:
#### Description
Only the password of the Facebook user is changed


tsvetomir-nedyalkov:
The FB account is locked again - will reopen the PR when I figure out how to fix it

PR #291, opened 2018-05-04
Skipping some of the MIC tests, as the used FB account gets locked

tsvetomir-nedyalkov:


tsvetomir-nedyalkov:
The FB account will be fixed in another PR

PR #301, opened 2018-05-25
MLIBZ-2455: TypeScript Definitions wrong for User.lookup()

thomasconner:
#### Description
The TypeScript definitions for `User.lookup()` function state that the function returns a `Promise` when it really returns an `Observable`.

#### Changes
- Update the TypeScript definitions file to state that the `User.lookup()` function returns an `Observable`.

#### Tests
- No testing needed. This was a documentation change.


vinaygahlawat:
LGTM. Is there an associated change that has to be made to the API reference guide in DevCenter?

thomasconner:
No there is not. The [DevCenter doc](https://devcenter.kinvey.com/nativescript/guides/users#lookup) is correct. This just corrects the TypeScript definitions file.

PR #302, opened 2018-05-29
MLIBZ-2323: Thrown error in onNext() in observable causes silent observable unsubscribe

thomasconner:
#### Description
When an error in the [onNext() function](https://devcenter.kinvey.com/html5/guides/datastore#Fetching) is thrown, the error is silently handled by the SDK and an unsubscribe from the observable is called, thus leaving the developer without notification that something went wrong.

A running sample describing the above can be found [here](https://dojo.telerik.com/AcEhE/5). Without a try-catch block, trying to access a property from the returned entities will cause an error as the first fetched result would be from the localStorage and thus an empty array. The thrown error will not be shown to the developer causing a silent unsubscribe from the observable.

#### Changes
- Don't swallow errors caused in the `onNext()` function for an Observable

#### Tests
- Added test to check that an error is not swallowed when created in the `onNext()` function


src/core/observable.js
```diff
@@ -128,7 +128,7 @@ class SafeSubscriber extends Subscriber {
     try {
       fn.call(this._context, value);
     } catch (err) {
-      this.unsubscribe();
+      this.error(err);
```
vinaygahlawat: Makes sense to throw an error here. But are there certain cases where we should still unsubscribe, or is this now something the developer should handle completely?

thomasconner: It will still unsubscribe after the error callback is called.

PR #305, opened 2018-06-12
MLIBZ-2366: Fix Google Auth on MIC for NativeScript SDK

thomasconner:
#### Description
MIC login with Google fails on NativeScript. The reason is an unsupported `user-agent`.

#### Changes
- Use `SFSafariViewController` on iOS for showing a MIC popup. Use `CustomTabs` on Android for showing a MIC popup.


packages/kinvey-nativescript-sdk/src/bundle.ts
```diff
@@ -29,7 +29,6 @@ export {
   Log,
   Metadata,
   Query,
-  User,
```
vinaygahlawat: Why is `User` removed here?

thomasconner: It was removed on this line because the `User` module is now exported by the [line](https://github.com/Kinvey/js-sdk/pull/305/files/b1f5ab460d683c7c6160e6b8d018098d2750a5c2#diff-10d68474809f3791aaa0c9fbeca55222R75). The `User` module is now extended by the NativeScript package.

packages/kinvey-nativescript-sdk/lib/before-checkForChanges.js
```diff
@@ -0,0 +1,44 @@
...
+  const appDirectoryPath = hookArgs && hookArgs.checkForChangesOpts && hookArgs.checkForChangesOpts.projectData && hookArgs.checkForChangesOpts.projectData.appDirectoryPath;
+
+  if (!appDirectoryPath) {
+    throw new Error('Unable to get path to app directory');
+  }
+
+  const pathToPackageJson = path.join(appDirectoryPath, 'package.json');
+  const packageJsonContent = JSON.parse(fs.readFileSync(pathToPackageJson));
+  const kinveyData = packageJsonContent.pluginsData && packageJsonContent.pluginsData[pkg.name];
+  const redirectUri = (kinveyData && kinveyData.config && kinveyData.config.redirectUri) || 'enterpriseauth://';
```
rosen-vladimirov: Do we want to have a default here? In case there's no redirectUri in the package.json, the plugin's AndroidManifest.xml and Info.plist will register a scheme for `enterpriseauth`. In case you apply this for two applications, we will have the undefined behavior where multiple apps register the same scheme. 
As a safer solution, I suggest directly throwing an error in case there's no redirectUri in the package.json:
```
if (!redirectUri) {
    throw new Error(`Unable to find redirectUri info in ${pathToPackageJson}. Please provide correct redirectUri`);
}
```

packages/kinvey-nativescript-sdk/package.json
```diff
@@ -36,15 +40,26 @@
     "push.d.ts",
     "push.ios.js",
     "push.ios.js.map",
+    "lib/before-prepare.js",
```
Martin-Hristov: The file's actual name is before-checkForChanges.js

thomasconner: Fixed

packages/kinvey-nativescript-sdk/lib/before-checkForChanges.js
thomasconner: Fixed

vinaygahlawat:
@rosen-vladimirov @Martin-Hristov - Do you have any more questions or concerns regarding this PR, or can this be approved and merged for release?

PR #309, opened 2018-06-15
MLIBZ-2321: Improve API reference documentation

thomasconner:
#### Description
Update API reference documentation comments.

#### Changes
- Add script to generate API reference docs.
- Update comments that generate API reference docs.


src/core/files.js
```diff
@@ -58,26 +61,10 @@ export class FileStore {
...
-   * files.find(query, {
-   *   tls: true, // Use transport layer security
-   *   ttl: 60 * 60 * 24, // 1 day in seconds
-   *   download: true // download the files
-   * }).then(function(files) {
-   *   ...
-   * }).catch(function(err) {
-   *   ...
-   * });
+   * @param   {boolean}               [options.tls=true]                        Use transport layer security
```
tejasranade: A good API reference will describe what the developer should expect when they choose different values for a property, particularly a boolean flag. So for example, you could say - 
```
By default, all requests to retrieve files are made with https.
By setting this flag to false, files are retrieved over unsecure http connections. This is not recommended because ...
```
I've provided a generic example which may not be accurate here, but the idea is to describe what the parameter *does* rather than just describe what the name *stands for*

src/core/files.js
```diff
@@ -58,26 +61,10 @@ export class FileStore {
...
+   * @param   {number}                [options.ttl]                             Specify a time to live for the _downloadURL
+   * @param   {boolean}               [options.download=false]                  Download the files
```
tejasranade: Similar to the TLS comment about. Needs some detail about why files aren't downloaded by default, where to find them if they are downloaded, etc.

src/core/acl.js
```diff
@@ -6,47 +6,59 @@ import { isDefined } from './utils';
...
- * var entity = { _acl: {} };
- * var acl = new Kinvey.Acl(entity);
  */
 export class Acl {
   constructor(entity) {
     if (isPlainObject(entity) === false) {
       throw new KinveyError('entity argument must be an object');
     }
 
+    entity._acl = entity._acl || {};
```
tejasranade: Even though this is a tiny (and harmless) change, generally it's bad form to make unrelated code changes in a docs PR.

src/core/datastore/datastore.doc.js
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable */
```
tejasranade: What is the motivation behind separating this in a `.doc.js` file? Do we have such files for other classes?

src/core/acl.js
thomasconner: I agree. This was not meant to be in this PR. I will remove it.

src/core/datastore/datastore.doc.js
thomasconner: The motivation is that in the DevCenter DataStore guides we discuss `DataStoreType` and not different data store classes. If I was to document the classes, it would be confusing since these classes are not available publicly. By using `datastore.doc.js` I can document a data store without requiring the implementation to fit into the way esdoc needs it to be to follow the way we discuss it in our guides.

This is the only area in the SDK where this needs to be done at the moment and we do not had a `.doc.js` anywhere else for other classes. This is the pattern would probably use if this scenario was encountered elsewhere ins the SDK.

src/core/files.js
thomasconner: Good point. I will make the change.

src/core/files.js
thomasconner: Good point. I will make the change.

PR #310, opened 2018-06-21
MLIBZ-2526: Remove _kmd.local

thomasconner:
#### Description
When an entity is created offline it will set the property `_kmd.local` to `true`. This needs to be removed before sending it to the backend. Currently it is not resulting in the following error:

```
{
  "error":"DataLinkConnectorError",
  "debug":"Unable to insert record. Error message = {\"_errors\":[{\"_errorMsg\":\"ERROR condition: An incomplete client request. (8020) (7211)\",\"_errorNum\":8020}]}",
  "description":"The DataLink request did not complete. See debug message for details."
}
```

#### Changes
- Remove `_kmd.local`

#### Tests
- Added unit test to verify `_kmd.local` is removed

### DIFF
--- a/src/core/datastore/sync/sync-manager.js
+++ b/src/core/datastore/sync/sync-manager.js
@@ -236,6 +236,9 @@ export class SyncManager {
   _sanitizeOfflineEntity(offlineEntity) {
     const copy = clone(offlineEntity);
     delete copy._id;
+    if (copy._kmd) {
+      delete copy._kmd.local;
+    }
     return copy;
   }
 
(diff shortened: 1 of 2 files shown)


src/core/datastore/sync/sync-manager.js
```diff
@@ -236,6 +236,9 @@ export class SyncManager {
   _sanitizeOfflineEntity(offlineEntity) {
     const copy = clone(offlineEntity);
     delete copy._id;
+    if (copy._kmd) {
+      delete copy._kmd.local;
```
heyzooi: what happens if `copy._kmd.local` is `null`?

PR #321, opened 2018-07-06
MLIBZ-2585: Queue requests when in the middle of refreshing a MIC access token

thomasconner:
#### Description
A response to an API request can result in a `401` which will attempt to refresh a MIC access token. When multiple requests are made in parallel this would cause multiple attempts to refresh the MIC access token. This caused an invalid state that was unrecoverable.

This PR only allows one attempt to refresh a MIC access token while queuing up any other API request made while the refresh attempt is made. Once the refresh attempt completes then the queued requests will be retried.

#### Changes
- Queue up pending API requests when refreshing a MIC access token

### DIFF
--- a/src/core/request/network.js
+++ b/src/core/request/network.js
@@ -190,26 +190,6 @@ function logout(origRequest) {
 
   return promise
     .then(() => {
-      // Logout the user because the refresh token request failed
-      const request = new KinveyRequest({
-        method: RequestMethod.POST,
-        authType: AuthType.Session,
-        url: url.format({
-          protocol: origRequest.client.apiProtocol,
-          host: origRequest.client.apiHost,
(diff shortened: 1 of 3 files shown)


src/core/request/network.js
```diff
@@ -523,8 +526,50 @@ export class KinveyRequest extends NetworkRequest {
...
+                    })
+                    .then(() => {
+                      return this.client.setActiveUser(null);
+                    })
+                    .catch((error) => {
+                      Log.error(error);
+                      return null;
+                    })
+                    .then(() => DataStore.clearCache({ client: this.client }))
+                    .catch((error) => {
```
vinaygahlawat: Will the `clearCache` call also clear the sync queue?

thomasconner: Yes it will.

PR #322, opened 2018-07-06
Fix TravisCI

thomasconner:


tsvetomir-nedyalkov:
@thomasconner  It seems that the Integration tests are not executed after these changes - I guess that the old .travis.yml will be reverted?

thomasconner:
This is on purpose. The integration tests import paths will have to be updated. There will be another PR towards the end of the refactor changes to updates the integration tests.

PR #323, opened 2018-07-06
MLIBZ-2590: Acl, Kmd, and Aggregation

thomasconner:


packages/kinvey-js-sdk/src/kmd.js
```diff
@@ -0,0 +1,111 @@
+import isPlainObject from 'lodash/isPlainObject';
+
```
vinaygahlawat: Similar question as with `acl`: if we are changing this class form `metadata` to `kmd`, is this a breaking change?

PR #333, opened 2018-07-23
MIBZ-2453: Kinvey.Files.downloadByUrl should call the error function if the server returns an error

thomasconner:
#### Description
**Steps To Reproduce:**

1. Stream a file with ttl = 0: 
```js
Kinvey.Files.download(fileId, { stream: true, ttl: 0 });
```

2. Try to download the file using the expired `__downloadURL_`: 
```js
Kinvey.Files.downloadByUrl(result._downloadURL);
```

**Expected Result:** The error function is called with the error from the server.
**Actual Result:** The success function is called with a string, containing an XML from the server:

```xml
<?xml version='1.0' encoding='UTF-8'?><Error><Code>ExpiredToken</Code><Message>The provided token has expired.</Message><Details>Request has expired: 1523979099</Details></Error>
```

#### Changes
- Reject with error from Network response.
- Only send X-Kinvey-Device-Information header to a kinvey.com endpoint.

### DIFF
--- a/src/html5/http.js
+++ b/src/html5/http.js
@@ -87,9 +87,12 @@ export class Html5HttpMiddleware extends Middleware {
         timeout,
         followRedirect
       } = request;
+      const kinveyUrlRegex = /kinvey\.com/gm;
 
-      // Add the X-Kinvey-Device-Information header
-      headers['X-Kinvey-Device-Information'] = deviceInformation(this.pkg);
+      if (kinveyUrlRegex.test(url)) {
+        // Add the X-Kinvey-Device-Information header
+        headers['X-Kinvey-Device-Information'] = deviceInformation(this.pkg);
+      }
 
       this.xhrRequest = xhr({
         method: method,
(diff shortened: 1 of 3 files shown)


src/html5/http.js
```diff
@@ -87,9 +87,12 @@ export class Html5HttpMiddleware extends Middleware {
         timeout,
         followRedirect
       } = request;
+      const kinveyUrlRegex = /kinvey\.com/gm;
 
-      // Add the X-Kinvey-Device-Information header
-      headers['X-Kinvey-Device-Information'] = deviceInformation(this.pkg);
+      if (kinveyUrlRegex.test(url)) {
+        // Add the X-Kinvey-Device-Information header
+        headers['X-Kinvey-Device-Information'] = deviceInformation(this.pkg);
```
vinaygahlawat: How is the setting of the `X-Kinvey-Device-Information` header related to the issue of calling success when an error is returned from the backend?

thomasconner: I was getting errors where Google was reject the `GET` request because of the `X-Kinvey-Device-Information` header. It was not causing the issue but it was preventing me from testing. I went ahead and made the change here.

PR #335, opened 2018-07-24
MLIBZ-2626: Fix TypeScript definitions for NativeScript Push

thomasconner:


thomasconner:
For Angular2+ the `Push` module is included in the package. For NativeScript, the `Push` module is separated. This is why there are 2 TypeScript definitions files.

PR #338, opened 2018-07-25
MLIBZ-2629:  The sdk does not work with Android 5.1 or older on PhoneGap

thomasconner:
#### Description
**Steps To Reproduce**:
1. Try to run the following test app on an Android 5.1 emulator: https://www.dropbox.com/s/882xmqccvhhxist/KinveyCordovaTestApp.zip?dl=1

**Expected Result**: No errors
**Actual Result**: The following error:

```js
Uncaught SyntaxError: Unexpected strict mode reserved word
```

It seems that the issue is in the following function:

```js
function lowerBound(array, value, comp) {
	let first = 0;
	let count = array.length;

	while (count > 0) {
		const step = (count / 2) | 0;
		let it = first + step;

		if (comp(array[it], value) <= 0) {
			first = ++it;
			count -= step + 1;
		} else {
			count = step;
		}
	}

	return first;
}
```

#### Changes
- Replace `p-queue` with `promise-queue`


package.json
```diff
@@ -14,7 +14,7 @@
     "local-storage-fallback": "~4.0.0",
     "lodash": "~4.17.4",
     "loglevel": "~1.5.1",
-    "p-queue": "^2.4.2",
+    "promise-queue": "github:thomasconner/promise-queue",
```
vinaygahlawat: Why is this version forked?

thomasconner: `p-queue` allowed us to pause and start the queue. Since it is not able to be used due to issues on PhoneGap we need to replace it with a similar package called `promise-queue`. `promise-queue` does not support pausing and starting of the queue so a fork was made to add the functionality

PR #343, opened 2018-08-03
MLIBZ-2466: Add new device info header information

thomasconner:
#### Description
Implements new `X-Kinvey-Device-Info` header spec.

#### Changes
- Adds `X-Kinvey-Device-Info` header to outgoing network requests sent to Kinvey.


src/node/http.js
```diff
@@ -21,6 +22,16 @@ function deviceInformation(pkg) {
   }).join(' ');
 }
 
+function deviceInformation2(pkg) {
+  return {
+    hv: 1,
+    os: os.platform(),
+    ov: os.release(),
+    sdk: pkg.name,
+    pv: os.release()
```
vinaygahlawat: Why are the `OS Version` and the `Platform Version` the same value here?

thomasconner: This was a mistake. I have fixed it.

thomasconner:
I had to make a change to make sure the `AndroidManifest.xml` is included when the `kinvey-nativescript-sdk` package is installed. Can you review and approve if everything looks good @vinaygahlawat?

PR #351, opened 2018-08-29
MLIBZ-2649: Use encryptionKey to encrypt SQLite database on NativeScript

thomasconner:
#### Description
Add encryption for SQLite on NativeScript.

#### Changes
- Use `encryptionKey` to open the SQLite database.

### DIFF
--- a/packages/kinvey-nativescript-sdk/package.json
+++ b/packages/kinvey-nativescript-sdk/package.json
@@ -9,7 +9,7 @@
   "main": "kinvey-nativescript-sdk.js",
   "typings": "kinvey.d.ts",
   "dependencies": {
-    "nativescript-sqlite": "~1.1.11",
+    "nativescript-sqlite": "latest",
     "rxjs": "~5.5.5"
   },
   "scripts": {
--- a/src/nativescript/nativescript-sql-module.ts
+++ b/src/nativescript/nativescript-sql-module.ts
@@ -7,9 +7,11 @@ const NativeScriptSQLite = require('nativescript-sqlite');
 
 export class NativescriptSqlModule {
   private _databaseName: string;
+  private _encryptionKey: string;
 
-  constructor(databaseName) {
+  constructor(databaseName, encryptionKey) {
     this._databaseName = databaseName;
+    this._encryptionKey = encryptionKey;
   }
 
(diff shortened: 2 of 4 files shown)


packages/kinvey-nativescript-sdk/package.json
```diff
@@ -9,7 +9,7 @@
   "main": "kinvey-nativescript-sdk.js",
   "typings": "kinvey.d.ts",
   "dependencies": {
-    "nativescript-sqlite": "~1.1.11",
+    "nativescript-sqlite": "latest",
```
vinaygahlawat: Do we use the `latest` for other dependencies? Why is this dependency different than others in terms of what version to use?

src/nativescript/nativescript-sql-module.ts
```diff
@@ -7,9 +7,11 @@ const NativeScriptSQLite = require('nativescript-sqlite');
 
 export class NativescriptSqlModule {
   private _databaseName: string;
+  private _encryptionKey: string;
 
-  constructor(databaseName) {
+  constructor(databaseName, encryptionKey) {
```
vinaygahlawat: Is this a breaking change, or does the module just create an unencrypted database if the `encyrptionKey` is null?

thomasconner: This is not a breaking change. An unencrypted database will be created if `encryptionKey` is `null`. I have tested this situation.

packages/kinvey-nativescript-sdk/package.json
vinaygahlawat: Also, since the latest version of `nativescript-sqlite` is now at v2, do we have to check for any breaking changes?

thomasconner: I am use the `latest` because this was what the README for using encryption said to do.

thomasconner: I have not noticed any changes that break how we are using `nativescript-sqlite`

thomasconner: v2 major change was in the way the plugin allowed other plugins to be loaded automatically such as the encryption plugin. There were no changes to the API for `nativescript-sqlite` according to the changelog

PR #361, opened 2018-09-28
MLIBZ-2543: Fix encoding issue when downloading files via the NativeScript SDK

thomasconner:
#### Changes
- Do not call `toString()` on the response data for file downloads.

### DIFF
--- a/src/nativescript/http.ts
+++ b/src/nativescript/http.ts
@@ -74,11 +74,16 @@ export class HttpMiddleware extends Middleware {
     };
     return (HttpRequest(options) as any)
       .then((response) => {
+        const contentType = response.headers['content-type'] || response.headers['Content-Type'];
         let data = response.content.raw;
 
-        try {
-          data = response.content.toString();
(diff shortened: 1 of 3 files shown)


src/nativescript/http.ts
```diff
@@ -74,11 +74,16 @@ export class HttpMiddleware extends Middleware {
...
         let data = response.content.raw;
 
-        try {
-          data = response.content.toString();
-        } catch (e) {}
+        if (contentType) {
+          if (contentType.indexOf('application/json') === 0 || contentType.indexOf('application/xml') === 0) {
+            try {
+              data = response.content.toString();
+            } catch (e) {}
```
vinaygahlawat: Does it make sense here to have a log message, so that we know for what situations we may be failing to call `toString()` when the content type is either `application/json` or `application/xml`?

thomasconner: I removed the check of the `Content-Type` header.

PR #364, opened 2018-10-04
MLIBZ-2671: Fix MIC on Microsoft Edge

thomasconner:
#### Description
`loginWithMIC()` uses a popup when on a web browser to display the identity providers login page. An interval is setup to poll this popup for url changes. Once the popup redirects after a successful login the interval will parse the code received from the redirect url and provide it to the SDK.

While the user is attempting to login, the url of the popup is on a different domain of the hosted application that initiated the login. This causes an error when checking the url of the popup that has to be handled and thrown out.

This change handles the error specific thrown by Microsoft Edge when trying to access the url for the popup.

#### Changes
- Handle the specific error thrown by Microsoft Edge for `loginWithMIC()`

#### Tests
- This was tested manually on Microsoft Edge running on a Windows 10 VM and Google Chrome. It is not possible at this time to automate this test. `loginWithMIC()` worked correctly after I made this change.

### DIFF
--- a/src/html5/popup.js
+++ b/src/html5/popup.js
@@ -17,7 +17,7 @@ export class Popup extends EventEmitter {
             this.emit('loadstart', event);
             this.emit('load', event);
           } catch (error) {
-            if (error.code !== global.DOMException.SECURITY_ERR) {
+            if (error.code !== global.DOMException.SECURITY_ERR && error.message.indexOf('Permission Denied') !== -1) {
               this.emit('error', error);
             }
           }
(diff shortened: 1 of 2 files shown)


src/html5/popup.js
```diff
@@ -17,7 +17,7 @@ export class Popup extends EventEmitter {
             this.emit('loadstart', event);
             this.emit('load', event);
           } catch (error) {
-            if (error.code !== global.DOMException.SECURITY_ERR) {
+            if (error.code !== global.DOMException.SECURITY_ERR && error.message.indexOf('Permission Denied') !== -1) {
```
vinaygahlawat: Please add a comment to point out what portion of this check is specific to `Microsoft Edge`.

thomasconner: I added the comment

PR #377, opened 2018-10-25
Conform with NativeScript preview application

Mitko-Kerezov:
Conform Kinvey-sdk with NativeScript preview application. 
MIC login improvements.

Includes:
* **Improvement** Add before-preview-sync hook  - this hook will fire whenever a user scans a QR code displayed in the console from a `tns preview` command. It will warn the user that their application might not work as expected in case they do not have the {N} Preview app uri scheme
* **Improvement** Implement preview app check - kinvey {N} sdk now detects at runtime whether running inside {N} Preview app. If this is the case disregard any arguments passed to `loginWithMIC` and use the preview app scheme `nsplayresume` instead 
* **Fix(iOS):** Visualize login page in topmost `UIViewController` - prior to this the login page was displayed on the root `UIViewController`. Whenever the root `UIViewController` was obscured (for example user has another `UIViewController`  on top of the root one - e.g. a modal dialog) the login page would not show at all. iOS does not permit displaying items on invisible `UIViewController`s.
* **Fix(Android):** Ensure main activity is `singleTask` - ensure the main activity's launchMode is set to `singleTask` (meaning there may exist only a single instance of said activity throughout the whole Android system). This is needed because without it behavior is as follows: User clicks `loginWithMIC`, a login page is displayed in Chrome, User logs in and is redirected back to the application where the whole activity is re-created. This leads to potential unwanted behavior because instead of the login just continuing normally it leads to re-creation of the whole application

Ping @thomasconner for review

### DIFF
--- a/packages/kinvey-nativescript-sdk/lib/before-preview-sync.js
+++ b/packages/kinvey-nativescript-sdk/lib/before-preview-sync.js
@@ -1,3 +1,3 @@
 module.exports = function ($logger) {
-  $logger.warn("If you are using loginWithMIC ensure that you have nsplayresume:// as one of your Redirect URI's in order for Mobile Identity Connect login to work in the Preview app.");
+  $logger.warn("If you are using loginWithMIC() ensure that you have added nsplayresume:// as a Redirect URI to your Mobile Identity Connect configuration at https://console.kinvey.com in order for Mobile Identity Connect login to work in the Preview app.");
 };
--- a/packages/kinvey-nativescript-sdk/lib/before-preview-sync.js
+++ b/packages/kinvey-nativescript-sdk/lib/before-preview-sync.js
@@ -1,0 +1,3 @@
+module.exports = function ($logger) {
+  $logger.warn("If you are using loginWithMIC ensure that you have nsplay:// as one of your Redirect URI's in order for Mobile Identity Connect login to work in the Preview app.");
+};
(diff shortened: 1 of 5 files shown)


src/nativescript/user.js
```diff
@@ -1,8 +1,10 @@
 import { User as CoreUser } from '../core/user';
 import { getDataFromPackageJson } from './utils';
+import * as app from "application";
 
 export class User extends CoreUser {
   loginWithMIC(redirectUri, authorizationGrant, options) {
+    redirectUri = isInsidePreviewApp() ? "nsplayresume://" : redirectUri;
```
thomasconner: Is the `nsplayresume://` scheme being setup as a custom url scheme for the app on iOS and an `IntentFilter` on Android?

thomasconner: See https://github.com/Kinvey/js-sdk/blob/aab4721338b22a6c51d29b4246ea0fcb4d4d18aa/packages/kinvey-nativescript-sdk/platforms/ios/Info.default.plist#L12 for iOS and https://github.com/Kinvey/js-sdk/blob/aab4721338b22a6c51d29b4246ea0fcb4d4d18aa/packages/kinvey-nativescript-sdk/platforms/android/AndroidManifest.default.xml#L8 for Android

thomasconner: Are we able to call `isInsidePreviewApp()` during a hook instead of in the code?

thomasconner: Ok great!

I think my unfamiliarity with how the Preview App works is why I was asking about calling `isInsidePreviewApp()` in a hook.

I need to spend some time (today) playing with the preview app but what you have changed looks good to me.

I will approve it once I can play with the preview app with your changes.

packages/kinvey-nativescript-sdk/lib/before-preview-sync.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function ($logger) {
+  $logger.warn("If you are using loginWithMIC ensure that you have nsplayresume:// as one of your Redirect URI's in order for Mobile Identity Connect login to work in the Preview app.");
```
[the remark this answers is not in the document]
Mitko-Kerezov: Done

PR #380, opened 2018-12-04
MLIBZ-2698: Throw error if no id is provided to findById()

thomasconner:
#### Description
When you call `collection.findById()` without an `id` no request is sent, which is correct, but `undefined` is returned instead of an error.

#### Changes
- throw an error when an `id` is not provided to `findById()`

### DIFF
--- a/src/core/datastore/networkstore.spec.js
+++ b/src/core/datastore/networkstore.spec.js
@@ -144,6 +144,17 @@ describe('NetworkStore', () => {
   });
 
   describe('findById()', () => {
+    it('should throw an error if an id is not provided', () => {
+      const store = new NetworkStore(collection);
+      return store.findById().toPromise()
+        .then(() => {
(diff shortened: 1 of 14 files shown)


src/core/datastore/networkstore.spec.js
```diff
@@ -144,6 +144,17 @@ describe('NetworkStore', () => {
...
 
   describe('findById()', () => {
+    it('should throw an error if an id is not provided', () => {
+      const store = new NetworkStore(collection);
+      return store.findById().toPromise()
+        .then(() => {
+          throw new Error('This should not happen.')
+        })
+        .catch((error) => {
+          expect(error).toBeA(KinveyError);
```
vinaygahlawat: To be similar to other tests, should there be another check for error type? Something similar to:
`expect(error.message).toEqual('No id was provided. An id must be provided.');`

PR #382, opened 2018-12-04
MLIBZ-2708: Throw NoActiveUserError when no active user exists and me() is called

thomasconner:
#### Description
When `me()` is called and no active user exists `null` is returned and no `_me` request is sent instead of throwing a `NoActiveUserError`.

#### Changes
- Throw `NoActiveUserError` when no active user exists and `me()` is called

### DIFF
--- a/src/core/user/user.spec.js
+++ b/src/core/user/user.spec.js
@@ -7,7 +7,7 @@ import { Acl } from '../acl';
 import { Metadata } from '../metadata';
 import { User } from './user';
 import { randomString } from '../utils';
-import { ActiveUserError, InvalidCredentialsError, KinveyError } from '../errors';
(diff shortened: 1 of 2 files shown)


src/core/user/user.spec.js
```diff
@@ -866,6 +866,17 @@ describe('User', () => {
       const user = new User();
       return chai.expect(user.me()).to.eventually.be.rejected;
     });
+
+    it.only('should fail if there is no active user', () => {
+      return UserMock.logout()
+        .then(() => User.me())
+        .then(() => {
+          throw new Error('This should not fail');
```
vinaygahlawat: I know it's not really part of the test, but I think this error message should be more accurate. We do expect this test to throw an exception, please change the string to something like `This test should not successfully resolve, it should throw an error.`

PR #388, opened 2019-01-16
Adding tests for kinvey.init(), and some others

Martin-Hristov:


package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@babel/runtime": "7.2.0",
     "axios": "0.18.0",
+    "idb": "^3.0.2",
```
thomasconner: This should be added to the `devDependencies`

PR #390, opened 2019-02-13
MLIBZ-2847: Auto DataStore Type

thomasconner:
#### Description
This PR removes the use of Observables, removes the Cache DataStore type, and adds the Auto DataStore type


src/datastore/index.js
```diff
@@ -1,47 +1,36 @@
...
-import { NetworkStore } from './networkstore';
-import { CacheStore } from './cachestore';
+import { clear as _clear } from './cache';
+import NetworkStore from './networkstore';
+import { SyncStore } from './syncstore';
+import AutoStore from './autostore';
 
 export const DataStoreType = {
-  Cache: 'Cache',
+  Auto: 'Auto',
```
heyzooi: are we going to replace `Cache`? or deprecate and remove later?

src/datastore/sync.js
```diff
@@ -136,8 +133,8 @@ export class Sync {
   }
 
   async push(query, options) {
-    const network = new NetworkStore(this.collectionName);
-    const cache = new DataStoreCache(this.collectionName, this.tag);
+    const network = new Network(this.collectionName);
+    const cache = new Cache(this.collectionName, this.tag);
```
heyzooi: will this cause breaking changes?

vinaygahlawat: @heyzooi Yes, in the js-sdk, the move to `deprecate` the Cache store is a breaking change, largely because the `find()` API is removing its use of observables.

src/datastore/index.js
vinaygahlawat: @heyzooi I don't think there is a clean way to deprecate methods in JS. In the js-sdk, the move to "deprecate" the Cache store is a breaking change, largely because the `find()` API is removing its use of observables. So since we cannot call the method obsolete and remove it at a later time, we should just go ahead and remove the `Cache` store altogether.

heyzooi: oh I see, make sense

heyzooi:
@thomasconner LGTM overall, the code looks good but worth to take a look why the unit tests are failing before merge

src/datastore/sync.js
thomasconner: No this will not. This is all internal.

PR #392, opened 2019-02-18
MLIBZ-2879: Push Notifications

thomasconner:
#### Description
Add support for push notifications on NativeScript.


src/push/index.nativescript.ts
```diff
@@ -0,0 +1,69 @@
+import { messaging, Message } from 'nativescript-plugin-firebase/messaging';
+import { device } from 'tns-core-modules/platform';
+import { formatKinveyUrl } from '../http/utils';
+import { KinveyRequest, RequestMethod } from '../http/request';
+import { Auth } from '../http/auth';
+import { get as getConfig } from '../kinvey/config';
+
+const PUSH_NAMESPACE = 'push';
+
+export async function register(callback: (message: Message) => void, options: any = {}) {
```
thomasconner: After creating the PR I realized I should change the callback parameter to be more generic. By making it more generic it makes the code sharing story easier. What do you think?

packages/kinvey-nativescript-sdk/package.json
```diff
@@ -24,6 +24,7 @@
     "loglevel": "~1.6.1",
     "loglevel-plugin-prefix": "~0.8.4",
     "lokijs": "~1.5.6",
+    "nativescript-plugin-firebase": "~7.7.0",
```
vinaygahlawat: Is there a package we used to previously use for `push` that we should now be removing?

packages/kinvey-nativescript-sdk/lib/before-checkForChanges.js
```diff
@@ -46,6 +46,13 @@ module.exports = function (hookArgs) {
           if (currentContent !== micInfoPlistFileContent) {
             fs.writeFileSync(destinationInfoPlistFile, micInfoPlistFileContent);
           }
+        } else {
+          const defaultInfoPlistFile = path.join(__dirname, '..', 'platforms', 'ios', 'Info.default.plist');
```
vinaygahlawat: Not sure if capitalization matters, but the file at the end of the path is `Info.default.plist`, but the actual filename is `info.default.plist`. May be worth changing it to be the same for both.

src/push/index.nativescript.ts
```diff
@@ -0,0 +1,69 @@
...
+  // Register for push notifications
+  const {
+    // Whether you want this plugin to automatically display the notifications or just notify the callback. Currently used on iOS only. Default true.
+    showNotifications = true,
+    // Whether you want this plugin to always handle the notifications when the app is in foreground. Currently used on iOS only. Default false.
+    showNotificationsWhenInForeground = true
```
vinaygahlawat: The comment says default is `false`, but the code has it set to `true`.

src/index.js
```diff
@@ -48,7 +49,10 @@ const SDK = {
   AuthorizationGrant: User.AuthorizationGrant,
 
   // Ping
-  ping
```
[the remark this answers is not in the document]
thomasconner: Good question. This is because the API that is documented and what we had been doing is `Kinvey.ping()` vs the other stuff is `Kinvey.User.login()` etc.

packages/kinvey-nativescript-sdk/package.json
thomasconner: Not for the refactor.

src/push/index.nativescript.ts
thomasconner: The default value for the plugin is false but I think we should change it to default to true for our SDK. I left the comment to read the default value for the plugin. Maybe I should reword the comment to be less confusing.

thomasconner: This has been fixed.

packages/kinvey-nativescript-sdk/lib/before-checkForChanges.js
thomasconner: Good catch. This has been fixed.

src/push/index.nativescript.ts
thomasconner: It is a little different for the existing code. Right now you call `Kinvey.Push.register()` to register the device with Kinvey. You would not pass your callback to the register call. Instead you would use `Kinvey.Push.onNotification()` to register your callback. The above is changing that to do it in one step. However, the callback signature is the same in that it will receive one parameter that contains the data for the push notification.

thomasconner: See https://github.com/Kinvey/js-sdk/blob/master/packages/kinvey-nativescript-sdk/src/push/common.ts#L26.

thomasconner: I went ahead and made the callback parameter type more generic for now.

PR #403, opened 2019-03-29
Add support for the new KS preview app scheme

tsvetie:
There are two preview apps - NS and Kinvey Studio, and each has its own scheme. Add support for the Kinvey Studio preview app.

### DIFF
--- a/src/nativescript/user.js
+++ b/src/nativescript/user.js
@@ -16,7 +16,7 @@ export class User extends CoreUser {
   }
 }
 
-function getRedirectUri() {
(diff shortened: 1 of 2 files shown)


src/nativescript/user.js
```diff
@@ -15,8 +16,17 @@ export class User extends CoreUser {
   }
 }
 
-function isInsidePreviewApp() {
-  return getAppIdentifier() === "org.nativescript.preview";
+function getRedirectUri() {
```
thomasconner: This function is missing the `redirectUri` parameter.

PR #409, opened 2019-05-09
MLIBZ-2848: Add new datastore type auto

thomasconner:
#### Description
Adding a new datastore type called `DataStoreType.Auto`.

For read operations, the datastore will try to read the data from the backend. If the network request to the backend fails, it will return whatever data is available in the local cache. if the network request to the backend succeeds, the local cache will be updated with the results.

For write operations, this datastore type will write to the local cache first and then push the items from the sync queue to the backend. This is the same behavior as `DataStoreType.Cache`.

#### Changes
- Deprecating the existing `DataStoreType.Cache` and implementing the new `DataStoreType.Auto`

#### Tests
- Added integration tests outlined in the [test plan](https://progresssoftware-my.sharepoint.com/:x:/r/personal/mhristov_progress_com/_layouts/15/Doc.aspx?sourcedoc=%7B6f13e99d-4648-4416-820d-fac0ca9bee8f%7D&action=default&gad=730)


tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  before(function () {
+    // Remove any existing data in the collection
+    return cleanUpAppData(collectionName);
+  });
+
+  before(function() {
+    // Signup an anonymous user
+    return User.signup();
```
Martin-Hristov: If we record the userId we can pass it to cleanUpAppData where it would be deleted and we would avoid the explicit deletion of the user in the '**after**' hook below.

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      .get(`/appdata/${process.env.APP_KEY}/${collectionName}`)
+      .replyWithError({
+        message: 'no network connection'
+      });
+
+    const collection = DataStore.collection(collectionName, DataStoreType.Auto);
+    const docs = await collection.find();
+    expect(docs).to.deep.equal(cachedDocs);
+    expect(scope.isDone()).to.equal(true);
+  });
```
Martin-Hristov: I think we should add at least a few more tests for this iteration of the feature - an error is returned when there is an actual problem non-connection related (no permission for example) and one for the happy path for **save**, **create, update, remove, findById, removeById** operations. For these operations, also a test where connection error appears and we make sure that the operations are reflected in the offline provider and the kinvey_sync. Many of the tests can easily be added if the appropriate suites are extended a bit. For example **common-crud.spec.js** has a **dataStoreTypes** array, which can be extended with the AutoStore. Of course behavior for AS would need to be tweaked in some tests, but this should save a lot of work and time.

tests/integration/specs/utils.js
```diff
@@ -185,7 +185,7 @@ export function validateEntity(dataStoreType, collectionName, expectedEntity, se
 export function cleanUpCollectionData(collectionName) {
   const networkStore = Kinvey.DataStore.collection(collectionName, Kinvey.DataStoreType.Network);
   const syncStore = Kinvey.DataStore.collection(collectionName, Kinvey.DataStoreType.Sync);
-  return networkStore.find().toPromise()
+  return networkStore.find()
```
Martin-Hristov: Does find now return an object for all stores and would code expecting observable still work?

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    return nock.restore();
+  });
+
+  before(function() {
+    // Initialize the SDK
+    return init({
+      appKey: process.env.APP_KEY,
+      appSecret: process.env.APP_SECRET
+    });
+  });
```
Martin-Hristov: It might be a good idea if we added the option to run the tests for various offline storage providers. I use the utils.setOfflineProvider() function in other suites:
```
before(() => {
      const initProperties = {
        appKey: process.env.APP_KEY,
        appSecret: process.env.APP_SECRET,
        masterSecret: process.env.MASTER_SECRET
      }
      return Kinvey.init(utilities.setOfflineProvider(initProperties, process.env.OFFLINE_STORAGE));
    });
```

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+  before(function() {
+    // Signup an anonymous user
+    return User.signup();
+  });
+
+  before(function() {
+    const collection = DataStore.collection(collectionName, DataStoreType.Sync);
+    return collection.save({ title: 'Cached Book' })
```
Martin-Hristov: If this hook is to be used for other tests - happy path for find() for example, it would be better if AutoStore is used

packages/js-sdk/src/datastore/index.ts
```diff
@@ -1,17 +1,20 @@
...
+import { AutoStore } from './autostore';
 
 export enum DataStoreType {
+  Auto = 'Auto',
   Cache = 'Cache',
   Network = 'Network',
   Sync = 'Sync'
 };
 
-export function collection(collectionName: string, type = DataStoreType.Cache, options: any = {}) {
```
vinaygahlawat: Is there a way to deprecate this `collection` function? If not, the default DataStoreType should remain `Cache`, to avoid making this release a breaking change.

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,1147 @@
...
+      try {
+        await syncTypeCollection.findById(sampleDocs[0]._id).toPromise();
+        throw new Error('This test should fail');
+      } catch (error) {
+        expect(error).to.be.instanceOf(Errors.NotFoundError);
+      }
+    });
+
+    it.skip('should delete items with tagged store', async function() {
+      const tag = randomString();
```
Martin-Hristov: Is it failing?

tests/integration/node/package.json
```diff
@@ -5,7 +5,7 @@
   "version": "0.0.12",
   "scripts": {
     "pretest": "node scripts/test.js",
-    "test": "mocha --opts mocha.opts \"test/**/*.spec.js\""
+    "test": "mocha --opts mocha.opts \"test/common/autostore.spec.js\""
```
Martin-Hristov: Is this temporary?

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,1147 @@
...
+      // Query
+      const query = new Query()
+        .contains(propertyName, [syncSampleDocs[0][propertyName], syncSampleDocs[1][propertyName], networkSampleDocs[0][propertyName]]);
+
+      // Sync
+      const { push, pull } = await autoTypeCollection.sync(query);
+      expect(push.length).to.equal(syncSampleDocs.length);
+      expect(pull).to.equal(3);
+
+      // TODO: Find SyncStore
```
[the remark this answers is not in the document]
thomasconner: This was fixed with https://github.com/Kinvey/js-sdk/pull/409/commits/a8bc722e8fe566ff01b049ad7772f39eeac683d9

tests/integration/node/package.json
thomasconner: This was fixed with https://github.com/Kinvey/js-sdk/pull/409/commits/04da399a69301965b65e950a33ae05133c91b959

tests/integration/specs/common/autostore.spec.js
thomasconner: https://github.com/Kinvey/js-sdk/pull/409/commits/04da399a69301965b65e950a33ae05133c91b959

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,1147 @@
...
+
+    it('should throw an error for trying to update without supplying and _id');
+
+    it('should create an item whose _id does not exist', async function() {
+      const autoTypeCollection = DataStore.collection(collectionName, DataStoreType.Auto);
+      const syncTypeCollection = DataStore.collection(collectionName, DataStoreType.Sync);
+      const networkTypeCollection = DataStore.collection(collectionName, DataStoreType.Network);
+
+      // Update a doc
+      const doc = await autoTypeCollection.update({ _id: randomString() });
```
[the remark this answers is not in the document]
thomasconner: https://github.com/Kinvey/js-sdk/pull/409/commits/04da399a69301965b65e950a33ae05133c91b959

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,1147 @@
...
+      const networkTypeCollection = DataStore.collection(collectionName, DataStoreType.Network);
+
+      // Create a doc
+      const doc = await autoTypeCollection.create({ _id: randomString() });
+
+      // Find with Network
+      expect(await networkTypeCollection.findById(doc._id)).to.deep.equal(doc);
+    });
+
+    it.skip('should update an item with existing _id', async function() {
```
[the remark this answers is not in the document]
thomasconner: Yes. The backend api returns an error and doesn't seem to handle updated existing entities using the create api.

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,1147 @@
...
+      // Find with Network
+      expect(await networkTypeCollection.findById(doc._id)).to.deep.equal(doc);
+    });
+
+    it('should create an item using _id provided', async function() {
+      const autoTypeCollection = DataStore.collection(collectionName, DataStoreType.Auto);
+      const networkTypeCollection = DataStore.collection(collectionName, DataStoreType.Network);
+
+      // Create a doc
+      const doc = await autoTypeCollection.create({ _id: randomString() });
```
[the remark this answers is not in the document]
thomasconner: https://github.com/Kinvey/js-sdk/pull/409/commits/04da399a69301965b65e950a33ae05133c91b959

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,1147 @@
...
+
+      // Create sample data
+      const sampleDocs = await createSampleCollectionData(collectionName, 3);
+
+      // Pull
+      expect(await autoTypeCollection.pull()).to.equal(3);
+
+      // Remove docs
+      const removedDoc1 = sampleDocs.shift();
+      await networkTypeCollection.removeById(removedDoc1._id);
```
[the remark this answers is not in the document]
thomasconner: https://github.com/Kinvey/js-sdk/pull/409/commits/04da399a69301965b65e950a33ae05133c91b959

tests/integration/specs/common/autostore.spec.js
```diff
@@ -0,0 +1,1147 @@
...
+      const docs = await syncTypeCollection.find().toPromise();
+      expect(docs.length).to.equal(sampleDocs.length);
+      docs.map((doc) => {
+        const sampleDoc = sampleDocs.find((sampleDoc) => sampleDoc._id === doc._id);
+        expect(doc).to.deep.equal(sampleDoc);
+      });
+    });
+
+    it('should use autopagination when turned on', async function() {
+      const autoTypeCollection = DataStore.collection(collectionName, DataStoreType.Auto, { useAutoPagination: true });
```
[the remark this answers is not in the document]
thomasconner: https://github.com/Kinvey/js-sdk/pull/409/commits/04da399a69301965b65e950a33ae05133c91b959

tests/integration/specs/common/autostore.spec.js
thomasconner: This should be an enhancement we add to the tests

tests/integration/specs/utils.js
thomasconner: Initially we were going to remove the observable api but decided against this breaking change. Instead the `Auto` datastore type won't use observables since it a new api. This change has been reverted.

tests/integration/specs/common/autostore.spec.js
thomasconner: This has been changed to the recommendation above.

packages/js-sdk/src/datastore/index.ts
thomasconner: The default datastore type is not `Auto` and not `Cache`.

vinaygahlawat: As per our conversation, please revert this to avoid the breaking change for now.

thomasconner: This was reverted with https://github.com/Kinvey/js-sdk/pull/409/commits/23e80c88e8c118efb2c95e7c85bb2c243f0b08a6

tests/integration/specs/common/autostore.spec.js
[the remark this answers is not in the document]
Martin-Hristov: This one is my mistake, create uses POST, so a duplicate _id is forbidden. You can delete it - it is not a valid case

PR #414, opened 2019-06-05
MLIBZ-3016: Multi Record Insert

thomasconner:
#### Description
Adds support for creating an array of docs to using the Kinvey backend API.

### DIFF
--- a/packages/js-sdk/src/datastore/networkstore.ts
+++ b/packages/js-sdk/src/datastore/networkstore.ts
@@ -195,10 +195,28 @@ export class NetworkStore {
   }
 
   async create(docs: any, options: any = {}) {
+    const batchSize = 100;
     const apiVersion = getApiVersion();
 
     if (apiVersion !== 5 && isArray(docs)) {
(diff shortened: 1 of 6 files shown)


packages/js-sdk/src/datastore/networkstore.ts
```diff
@@ -192,8 +194,10 @@ export class NetworkStore {
     return stream;
   }
 
-  async create(doc: any, options: any = {}) {
-    if (isArray(doc)) {
+  async create(docs: any, options: any = {}) {
+    const apiVersion = getApiVersion();
+
+    if (apiVersion !== 5 && isArray(docs)) {
```
vinaygahlawat: Missing a check to make sure the number of items in the `docs` array is < 100.

thomasconner: I added a `batchCreate()` utility function to process an array of `docs` that is greater than 100 with commit https://github.com/Kinvey/js-sdk/pull/414/commits/41c88d98e9addbe66cbe6dbf59ff932624bf9228.

PR #421, opened 2019-06-20
MLIBZ-3061: Add http adapter to kinvey-react-native-sdk

thomasconner:


packages/react-native-sdk/src/httpAdapter.ts
```diff
@@ -0,0 +1,57 @@
...
+import { NetworkError } from 'kinvey-js-sdk/lib/errors/network';
+import { TimeoutError } from 'kinvey-js-sdk/lib/errors/timeout';
+import { name, version } from '../package.json';
+
+export function deviceInfo() {
+  return {
+    hv: 1,
+    os: Platform.OS,
+    ov: Platform.Version,
+    pv: Platform.Version,
```
vinaygahlawat: Should both _ov_ and _pv_ be set to _Platform.Version_?

PR #429, opened 2019-07-02
MLIBZ-2999: Monitor network connection and reconnect PubNub

thomasconner:
#### Description
Live services do not resume receiving updates after connection problems are resolve. This fixes that issue by monitoring the network connectivity and reconnecting `PubNub`.

#### Changes
- Monitor network connectivity and reconnect to `PubNub`

### DIFF
--- a/packages/nativescript-sdk/src/nativescript/init.ts
+++ b/packages/nativescript-sdk/src/nativescript/init.ts
@@ -5,7 +5,6 @@ import * as SessionStore from './sessionStore';
 import * as Popup from './popup';
 import { getStorageAdapter, StorageProvider } from './storage';
 import { getDataFromPackageJson } from './utils';
-import * as live from './live';
 
 export interface KinveyConfig {
   appKey: string;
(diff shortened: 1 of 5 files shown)


vinaygahlawat:
Do any automated tests currently exist that exercise Pubnub reconnection? If not, is it feasible to do so in integration tests? I understand it may be difficult to automate this, just curious.

thomasconner:
@vinaygahlawat No we do not any automated tests that test this scenario. Currently it is difficult to simulate network connectivity problems in integration tests.

packages/nativescript-sdk/src/nativescript/init.ts
```diff
@@ -26,6 +27,7 @@ export function init(config?: KinveyConfig) {
     storageAdapter: getStorageAdapter(mergedConfig.storage),
     pubnub: PubNub
   })
+  live.startMonitoring();
```
vinaygahlawat: Should we start monitoring live service at init time, or only when we have an active user?

thomasconner: Technically live service does not work without an active user but we have not overridden the `login()` function yet which we would need to in order to start monitoring the network. This is also slightly more complicated due to the many ways a user can login and the possibility that they also logout.

vinaygahlawat: As per our conversation, I think it would be best to tie the start/stop of monitoring to user login/logout, so we are not unnecessarily monitoring/initializing live service.

PR #430, opened 2019-07-12
MLIBZ-2975: Add validation to User.signup()

thomasconner:


packages/js-sdk/src/user/signup.ts
```diff
@@ -1,13 +1,19 @@
 import isEmpty from 'lodash/isEmpty';
+import isPlainObject from 'lodash/isPlainObject';
 import { ActiveUserError } from '../errors/activeUser';
+import { KinveyError } from '../errors/kinvey';
 import { setSession, formatKinveyBaasUrl, HttpRequestMethod, KinveyHttpRequest, KinveyBaasNamespace, KinveyHttpAuth } from '../http';
 import { getActiveUser } from './getActiveUser';
 import { User } from './user';
 
-export async function signup(data: object | User, options: { timeout?: number, state?: boolean } = {}) {
+export async function signup(data?: object | User, options: { timeout?: number, state?: boolean } = {}) {
```
heyzooi: do you have a version of signup where no parameter is required?

heyzooi:
nevermind I saw the other PR a figured out

packages/js-sdk/src/user/signup.ts
heyzooi: nevermind I saw the other PR a figured out

packages/js-sdk/src/user/signup.ts
```diff
@@ -1,13 +1,19 @@
...
   const activeUser = getActiveUser();
   const { state = true } = options;
 
+  if (data && !isPlainObject(data)) {
```
manoldonev: @thomasconner Not obvious but this essentially is a breaking change introduced with a patch version of kinvey-js-sdk library and as a cascading effect it is a breaking change in kinvey-nativescript-sdk as well. See https://github.com/NativeScript/nativescript-app-templates/pull/72

thomasconner: You are right that this is not that obvious and should have been a minor version bump. Thanks for pointing this out. We are working hard to make sure our releases follow the [Semantic Versioning](https://semver.org/) for our releases.

PR #454, opened 2019-08-05
Add Cirrus CI support

thomasconner:
#### Description
This PR adds the necessary changes to use [Cirrus CI](https://cirrus-ci.org).


thomasconner:
We haven't been using TravisCI for some time now due to the complexity in debugging issues we were having while using it. Why leave the config? We can always recover it if needed but I doubt we will need it anymore. We should remove it I think if we are going to move to another CI.

PR #471, opened 2019-08-16
MLIBZ-3044: Add offline storage to React Native SDK

thomasconner:
This adds support for storing data offline with the React Native SDK using [AsyncStorage](https://github.com/react-native-community/async-storage).


packages/react-native-sdk/package.json
```diff
@@ -31,12 +31,16 @@
...
+    "react-native": "^0.60.0"
   },
   "devDependencies": {
+    "@react-native-community/async-storage": "^1.6.1",
     "@types/react-native": "^0.60.3",
+    "@types/react-native-sqlite-storage": "^3.3.1",
     "del-cli": "2.0.0",
     "fs-extra": "~8.1.0",
+    "react-native-sqlite-storage": "^3.3.11",
     "typescript": "~3.5.3"
```
vinaygahlawat: If we are not using SQLite, we should remove these SQLite dependencies.

thomasconner: Yep I agree. I noticed it after I created the PR so I removed SQLite as a dependency with cb411fb.

PR #481, opened 2019-08-21
fix: getter is deprecated

bradmartin:
#### Description
utils.ios.getter was deprecated in NativeScript.

#### Changes
- Removed getter method and unused import.


bradmartin:
@thomasconner any chance of a bug release soon for the kinvey-nativescript-sdk? I believe the current published version is still using an old version of background-http and my logs are cluttered with `getter is deprecated` messages.

thomasconner:
@bradmartin Yes we are getting close to making a new release.

PR #554, opened 2020-02-25
KDEV-295: Allow multi-insert only via create function

alexanderfilipov:
Multi-insert should be supported only via `create` function. Calling `save` with multiple entities should not be possible.

### DIFF
--- a/packages/js-sdk/src/datastore/networkstore.ts
+++ b/packages/js-sdk/src/datastore/networkstore.ts
@@ -198,22 +198,28 @@ export class NetworkStore {
     const batchSize = 100;
     const apiVersion = getApiVersion();
 
-    if (apiVersion !== 5 && isArray(docs)) {
+    if (apiVersion < 5 && isArray(docs)) {
       throw new KinveyError('Unable to create an array of entities. Please create entities one by one.');
     }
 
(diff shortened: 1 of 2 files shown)


packages/js-sdk/src/datastore/networkstore.ts
```diff
@@ -270,16 +252,20 @@ export class NetworkStore {
     return response.data;
   }
 
-  save(docs: any, options?: any) {
-    if (!isArray(docs)) {
-      const kmd = new Kmd(cloneDeep(docs));
+  save(doc: any, options?: any) {
+    const apiVersion = getApiVersion();
+    if (apiVersion >= 5 && isArray(doc)) {
+      throw new KinveyError('Unable to save an array of entities. To insert multiple entities, please use "create" function.');
```
georgiwe: Nit:
Think calling `create` a method is better, since it's a method of a class, not a standalone function.

alexanderfilipov:
Brought back the code that splits the entities array in batches of 100. Turned out KCS has a configured limit of 100 objects after all.

alexanderfilipov:
Something wrong happened with this PR, I pushed some changes while github was experiencing problems. I will recreate it.

PR #555, opened 2020-02-27
KDEV-295: Allow multi-insert only via create function

alexanderfilipov:
Recreated #554 
Previous comments have been addressed.


packages/js-sdk/src/datastore/cachestore.ts
```diff
@@ -171,6 +172,36 @@ export class CacheStore {
...
+      throw new KinveyError('Unable to create an array of entities. Please create entities one by one or use API version 5 or newer.');
+    }
+
+    const createManyResult = {
+      entities: new Array(docs.length).fill(null),
+      errors: []
+    };
+
+    const createPromises = docs.map((doc, index) => {
+      return this.createOne(doc, options)
```
georgiwe: This could be pretty slow if things still work the way I remember them. There's a `createOne` but no `createMany`?

alexanderfilipov: Well, this doesn't improve things but doesn't make them worse either. This change is meant to unify the data store APIs, so all of them have a `create` method that accepts an array. There isn't much we can do for the cache store besides iterating the entities.

alexanderfilipov:
I've added `createMany` and I'm not planning more changes for the moment. Please have a final look.

packages/js-sdk/src/datastore/networkstore.ts
```diff
@@ -198,22 +198,28 @@ export class NetworkStore {
...
+          return previousResult;
         }
 
         const batch = docs.slice(i, i + batchSize);
-        i += batchSize;
         const result = await this.create(batch, options);
-        return batchCreate(createResults.concat(result));
+        result.errors.forEach((e) => (e.index += i));
+        const combinedResult = {
+          entities: previousResult.entities.concat(result.entities),
```
georgiwe: `docs.slice(i, i + batchSize)` would ideally be a loop, working with indices of the array, and not create slices off the array each time. Think it would be good to not add to this overhead by copying the entire results array for each batch. I mean `concat` creates a new array every time. I think it's better to do `previousResult.push(... result.entities)` - this way we only iterate over the batch, not all the results for every batch.

packages/js-sdk/src/datastore/cachestore.ts
```diff
@@ -145,9 +146,9 @@ export class CacheStore {
     return stream;
   }
 
-  async create(doc: any, options: any = {}) {
+  async createOne(doc: any, options: any = {}) {
```
georgiwe: Could annotate `createOne` and `createMany` with a `private` access modifier. Or maybe `protected` makes more sense, not sure what the inheritance model is nowadays. Think it also makes sense to keep the "underscore for non-public methods" since the transpiled code has no indication for end-users.

PR #568, opened 2020-07-10
KDEV-865: Catch general errors and aggregate them only for push function

alexanderfilipov:
The previous fix #567 should be applied only for `push` function, because it may perform different CRUD actions and should not stop on first error. For other operations like create or update, the general error should be thrown immediately.

### DIFF
--- a/packages/js-sdk/src/datastore/cachestore.ts
+++ b/packages/js-sdk/src/datastore/cachestore.ts
@@ -347,7 +347,7 @@ export class CacheStore {
   push(query?: Query, options: any = {}) {
     const sync = new Sync(this.collectionName, this.tag);
     // eslint-disable-next-line no-param-reassign
-    options.catchGenericErrors = true;
+    options.catchGeneralErrors = true;
     return sync.push(undefined, options);
   }
 
--- a/packages/js-sdk/src/datastore/cachestore.ts
+++ b/packages/js-sdk/src/datastore/cachestore.ts
@@ -344,8 +344,10 @@ export class CacheStore {
     return { count };
   }
 
-  push(query?: Query, options?: any) {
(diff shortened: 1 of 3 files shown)


packages/js-sdk/src/datastore/cachestore.ts
```diff
@@ -344,8 +344,10 @@ export class CacheStore {
     return { count };
   }
 
-  push(query?: Query, options?: any) {
+  push(query?: Query, options: any = {}) {
     const sync = new Sync(this.collectionName, this.tag);
+    // eslint-disable-next-line no-param-reassign
+    options.catchGenericErrors = true;
```
georgiwe, line 350: Maybe "general errors" is more appropriate?

packages/js-sdk/src/datastore/sync.ts
```diff
@@ -155,15 +155,19 @@ export class Sync {
...
-          totalPushResults.push({
-            _id: doc.entityId,
-            operation: SyncEvent.Create,
-            entity: entitiesForInsert[index],
-            error
+        // In case of a general batch insert error, do not break the push operation and aggregate all errors in the result
+        if (options.catchGenericErrors === true) {
+          return syncDocsForInsert.forEach((doc, index) => {
+            totalPushResults.push({
+              _id: doc.entityId,
```
georgiwe, line 162: There was some discussion about adding an `index` property - not sure what was decided on that.

alexanderfilipov: This should not actually happen anymore because `create` would get the general error and not the array of individual errors.

georgiwe: I thought it was an issue which concerned push() as well. Alright then.

PR #570, opened 2020-09-24
Feat: NativeScript 7 Support

bradmartin:
#### Description
Adds support for NativeScript 7

#### Changes
- update tsconfigs
- update imports
- update dependencies for dev and plugin

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "editor.formatOnSave": false,
+  "editor.codeActionsOnSave": {
+      "source.organizeImports": false
+  }
+}
--- a/packages/nativescript-sdk/package-lock.json
+++ b/packages/nativescript-sdk/package-lock.json
@@ -1,3921 +1,2312 @@
 {
-	"name": "kinvey-nativescript-sdk",
-	"version": "5.0.0",
-	"lockfileVersion": 1,
-	"requires": true,
-	"dependencies": {
-		"@angular/compiler": {
-			"version": "8.2.1",
-			"resolved": "https://registry.npmjs.org/@angular/compiler/-/compiler-8.2.1.tgz",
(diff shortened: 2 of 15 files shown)


bradmartin:
Leaving this as a draft because one plugin `nativescript-urlhandler` needs to merge a PR that adds NS7 support. https://github.com/hypery2k/nativescript-urlhandler/pull/116 It would be possible to just add a fork of that, but I'm unaware how the Kinvey team wants to handle that.

bradmartin:
I'm actually not even sure what the purpose of the `popup` is within this plugin for NativeScript 🤔

bradmartin:
I've published a working version under `@bradmartin/kinvey-nativescript-sdk` with some minor modifications to get this working for majority of use cases for NS7.

steinerj:
> I'm actually not even sure what the purpose of the `popup` is within this plugin for NativeScript 🤔

Not sure how the popup relates to the urlhandler, but the popup is used for Kinvey's federated login mechanism "MIC" - https://devcenter.kinvey.com/nativescript/guides/mobile-identity-connect

ventsislav-georgiev:
@bradmartin Is there a way to make the library compatible to be used with both NS7 and NS6?

bradmartin:
I don't believe so, NS7 was a breaking change, so it would require a major version change. Mainly for the ES2017 support set in the tsconfig. Also several imports were changed, which also goes in line with the ES2017 (full tree-shaking benefit) to use the new `@nativescript/core` package over the older `tns-core-modules`.

I suppose there might be a way to make it work with some approach, but I'd say the benefit isn't worth the effort or maintenance in the future.

ventsislav-georgiev:
@bradmartin the tests couldn't run from this branch so I moved it to: https://github.com/Kinvey/js-sdk/pull/574
Thank you for the PR!

PR #577, opened 2021-01-26
KDEV-1451: {N} SDK - Fix webpack failing to find firebase messaging package

ventsislav-georgiev:
Importing
```ts
import { firebase, messaging } from '@nativescript/firebase';
```
Is being compiled by webpack expecting `@nativescript/firebase/firebase.js` to have a `messaging` export, but there is no such export and the import fails.


bradmartin:
I just noticed I opened the same PR 👍 😄 thanks for this. Was throwing me for a loop earlier.

tsvetomir-nedyalkov (approved):
LGTM

PR #578, opened 2021-01-26
fix: messaging import was incorrect with NS7 firebase

bradmartin:
#### Description

```typescript
import { messaging } from '@nativescript/firebase';
```
will fail and `messaging` will be undefined. The fix it to use `@nativescript/firebase/messaging`

#### Changes

Only changed the import, and ordered the import order.


bradmartin:
PR addressing the issue was opened earlier today.

PR #581, opened 2021-03-12
KDEV-721: Use all Query instance features when doing an autoStore.find()

georgiwe:
#### Description
AutoStore now uses the specified paging settings (skip, limit) and projection, when performing a `.find()`. This will bring about a side effect that fetching pages in a non-sequential manner could result in discrepancies between what's been fetched and what is stored for offline use. We will be addressing this issue in a planned later redesign of the AutoStore.

#### Changes
- Extract the logic to be used for `.find()` in a separate method and preserve the current behaviour of `.pull()` and `CacheStore.find()`
- Use the old `AutoStore.find()` logic when using delta set, for the time being
- Fix a wrong validation for `AutoStore.group()` which prevented its use
- `AutoStore.find()` with a projection (the `fields` query option) fetches all fields from the server, stores them offline and then prunes unwated fields before returning the result. The reason is that the items can be used with different projections offline and to support updates without removing fields on the server.

### DIFF
--- a/packages/js-sdk/src/datastore/autostore.ts
+++ b/packages/js-sdk/src/datastore/autostore.ts
@@ -12,15 +12,19 @@ export class AutoStore extends CacheStore {
   }
 
   async find(query?: Query, options: any = {}) {
-    const cache = new DataStoreCache(this.collectionName, this.tag);
-
     if (query && !(query instanceof Query)) {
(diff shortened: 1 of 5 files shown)


alexanderfilipov (approved):
LGTM. Maybe it makes sense to bumb the minor version?

alexanderfilipov (approved):

PR #589, opened 2021-04-11
fix(cachestore): typeerror: reduce of empty array with no initial value

itamitegama:
paginationResults.reduce() called without an initial value fails when paginationResults is empty.

#### Description
autostore pull fails when no pagination is needed

#### Changes
add initial empty array

### DIFF
--- a/packages/js-sdk/src/datastore/cachestore.ts
+++ b/packages/js-sdk/src/datastore/cachestore.ts
@@ -483,9 +483,9 @@ export class CacheStore {
       await queryCache.save(queryCacheDoc);
 
       if (returnCounts) {
-        return paginationResults.reduce((result: number, current: number) => (result + current));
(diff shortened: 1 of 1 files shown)


packages/js-sdk/src/datastore/cachestore.ts
```diff
@@ -483,9 +483,9 @@ export class CacheStore {
       await queryCache.save(queryCacheDoc);
 
       if (returnCounts) {
-        return paginationResults.reduce((result: number, current: number) => (result + current));
+        return paginationResults.reduce((result: number, current: number) => (result + current), []);
```
georgiwe, line 486: This one should be `0`.

tsvetomir-nedyalkov:
@itamitegamat the fix is released with #590 - thank you!

PR #592, opened 2021-04-16
KDEV-1575: Improve the login process

DennyGD:
This PR adds two-factor authentication functionality to the SDK, makes changes to the existing login method and allows invalidating all of user's access tokens.
* Add `loginWithMFA` - sign in a user who has MFA enabled or a user who belongs to an app that has MFA required. API version (apiVersion) should be no less than 6. If the user trusts the device, the device token will be sent along with the username and password.
* Add `loginWithRecoveryCode` - sign in a user who has MFA enabled or a user who belongs to an app that has MFA required.
* Change the existing `login` method's signature to prevent passing socialIdentity. Now, it accepts just username and password.
* Add `invalidateTokens` to expose baas functionality for invalidating all access tokens.

**Note:** MFA tests are expected to fail because MFA backend functionality is not released yet.


packages/js-sdk/src/user/loginWithMFA.ts
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    if (mfaCompleteResult.trustDevice != null && !isBoolean(mfaCompleteResult.trustDevice)) {
+      throw new KinveyError(errMsgTrustDevice);
+    }
+    const trustDevice = mfaCompleteResult.trustDevice || false;
+    const mfaData = await executeCompleteRequest(mfaCompleteResult.code, trustDevice);
+    return mfaData;
+  } catch (err) {
+    if (err.message === errMsgNoCode || err.message === errMsgTrustDevice || !(err instanceof KinveyError)) {
```
ventsislav-georgiev, line 77: Moving mfaComplete outside of the try block will remove this check and have the same behaviour.

packages/js-sdk/src/user/user.ts
```diff
@@ -247,4 +247,29 @@ export class User {
 
     return this;
   }
+
+  async invalidateTokens() {
+    if (!this.isActive()) {
+      return this;
+    }
+
+    this.unregisterFromLiveService();
```
ventsislav-georgiev, line 256: The whole method seems to match the logout and is not trivial, better to reuse it and just extract the try/catch block to be passed as parameter.

packages/js-sdk/src/user/loginWithMFA.ts
```diff
@@ -0,0 +1,168 @@
...
+    throw new KinveyError(errMsgNoAuthenticators);
+  }
+
+  const context = {
+    mfaSessionToken: loginResult.mfaSessionToken,
+    retries: 0,
+    authenticator: null,
+  };
+  const selectedAuthenticatorId = await selectAuthenticator(loginResult.authenticators, context);
+  if (selectedAuthenticatorId == null) {
```
ventsislav-georgiev, line 137: Any non null result from the select will bypass the check. This check should be done after we find the authenticator from the authenticators list.

DennyGD: We care about the null/undefined scenario because this is the way to stop the login process. If the end user decides to abandon the login while selecting the authenticator, then the developer should return null.
If selectAuthenticator returns a non-existent ID, then the backend will return a NotFound error which I think will be a rare scenario. Maybe, it could happen during development, so I don't think we should handle it client-side.

ventsislav-georgiev: but we already know the authenticators from the loginResult here:
`context.authenticator = loginResult.authenticators.find((a) => a.id === selectedAuthenticatorId)`

my suggestion is simply to move the check after the above code and make it `if (!context.authenticator) { ... }` as we don't need to go to the backend in this scenario.

ventsislav-georgiev (approved):
LGTM

PR #595, opened 2021-05-27
KDEV-1629: Integration tests setup for react-native-sdk

alexanderfilipov:
Integration tests for `react-native` package run with `jest`. Changes:
* Updated most dependencies of the package to latest versions
* Adjusted the mocha-style test specs to be compatible with `jest`. For example, `describe`s should not have descriptions.
* Jest runs the test suites in parallel by default, but there seem to be dependencies between our suites, so they will be run sequentially for the time being. I've verified the tests pass locally. It looks like the current CI is broken though and we will have to fix that.
* The script used for copying the specs into the package-specific test folders was copy-pasted a couple of times, so I've moved it one level up and made it reusable by adding parameters.
* Adjusted the available unit tests for push notifications, so they pass, but we need to work more on the notifications support (separate task).


packages/react-native-sdk/package.json
```diff
@@ -29,43 +29,42 @@
     "test": "jest"
   },
   "dependencies": {
-    "axios": "0.19.0",
-    "kinvey-js-sdk": "^5.1.1",
-    "lodash": "^4.17.15",
-    "pubnub": "4.25.2"
+    "axios": "0.21.1",
+    "kinvey-js-sdk": "6.0.0-next.1",
```
ventsislav-georgiev, line 33: Is this set temporary for running the tests, before we release the next major?

packages/react-native-sdk/package.json
```diff
@@ -29,43 +29,42 @@
...
+    "lodash": "4.17.21",
+    "pubnub": "4.32.1"
   },
   "peerDependencies": {
-    "@react-native-community/async-storage": "^1.6.1",
-    "react-native": "^0.60.5",
-    "react-native-push-notification": "^3.1.9"
+    "@react-native-async-storage/async-storage": "1.15.4",
```
ventsislav-georgiev, line 38: It may be better to keep using the caret range for the peer dependency versions. Allowing users to update them without warnings.

packages/react-native-sdk/src/push.ts
```diff
@@ -21,22 +21,26 @@ async function registerDeviceWithKinvey(os: string, token: string): Promise<void
...
+  };
+
+  await new KinveyHttpRequest(requestOptions).execute();
 }
 
 export async function register(options: PushNotificationOptions = {}): Promise<string> {
   const promise = new Promise<string>((resolve, reject): void => {
     RNPushNotification.configure({
-      senderID: options.senderID,
+      // senderID: options.senderID,
```
ventsislav-georgiev, line 32: Let's delete this legacy property instead of commenting it

ventsislav-georgiev (approved):
After comments

packages/react-native-sdk/package.json
alexanderfilipov: Yes, exactly. We are going to release `6.0.0` core SDK with MFA and some other changes first, and then RN as a following version.

alexanderfilipov:
Closing in favour of https://github.com/Kinvey/js-sdk/pull/599 which also contains these and other react-related changes.

PR #598, opened 2021-06-04
KDEV-1752: Async SessionStore API

alexanderfilipov:
The SessionStore interface should expose async functions in order to be more flexible for future implementations for various platforms.

### DIFF
--- a/packages/html5-sdk/src/sessionStore.ts
+++ b/packages/html5-sdk/src/sessionStore.ts
@@ -1,13 +1,13 @@
-export function get(key: string) {
+export async function get(key: string): Promise<string> {
   return window.localStorage.getItem(key);
 }
 
-export function set(key: string, session: string) {
+export async function set(key: string, session: string): Promise<boolean> {
   window.localStorage.setItem(key, session);
   return true;
 }
 
-export function remove(key: string): boolean {
+export async function remove(key: string): Promise<boolean> {
   window.localStorage.removeItem(key);
   return true;
 }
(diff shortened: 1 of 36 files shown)


packages/html5-sdk/src/sessionStore.ts
```diff
@@ -1,13 +1,13 @@
-export function get(key: string) {
+export async function get(key: string): Promise<string> {
```
DennyGD, line 1: I don't think that it returns necessarily a string, more like `Promise<any>`

packages/js-sdk/src/http/session.ts
```diff
@@ -20,54 +20,55 @@ export function getKey() {
   return `${getAppKey()}.active_user`;
 }
 
-export function getSession(): SessionObject | null {
-  const session = getStore().get(getKey());
+export async function getSession(): Promise<SessionObject> {
```
DennyGD, line 23: It seems that it still might return `null` instead of `SessionObject`

packages/js-sdk/src/http/session.ts
```diff
@@ -20,54 +20,55 @@ export function getKey() {
...
+export async function removeMFASessionToken(): Promise<boolean> {
   return getStore().remove(getMFAKey());
 }
 
 function getDeviceTokenKey(username: string): string {
   return `${getAppKey()}.${username}.device_token`;
 }
 
-export function getDeviceToken(username: string): string | undefined {
+export async function getDeviceToken(username: string): Promise<string> {
```
DennyGD, line 59: Won't it return `undefined` in case there's no device token found?

DennyGD (approved):

PR #599, opened 2021-06-09
KDEV-1629, KDEV-1630, KDEV-1631: Secure session store and MIC login for ReactNative

alexanderfilipov:
* Setup integration tests for ReactNative
* Secure session store implementation
* MIC login with popup browser


packages/react-native-sdk/package.json
```diff
@@ -37,6 +37,7 @@
   "peerDependencies": {
     "@react-native-async-storage/async-storage": "1.15.4",
     "react-native": "0.64.1",
+    "react-native-keychain": "7.0.0",
```
ventsislav-georgiev, line 40: Better to use the caret range instead of the exact version.

ventsislav-georgiev (approved):

PR #600, opened 2021-06-09
Add libxshmfence1 library to CI environment for HTML5 tests

alexanderfilipov:
HTML5 integration tests were failing with:
```
Cannot start Chrome
/tmp/cirrus-ci-build/node_modules/chromium/lib/chromium/chrome-linux/chrome: error while loading shared libraries: libxshmfence.so.1: cannot open shared object file: No such file or directory
```


ventsislav-georgiev (approved):
LGTM

PR #607, opened 2021-06-18
KDEV-1770: Fix the angular build

DennyGD:
* The problem was that there were references to the (core) js sdk in the angular sdk. These references led to the changed folder structure. The interfaces are now exported from the js sdk and then from the html5 sdk to the angular sdk.


alexanderfilipov (approved):
LGTM

PR #609, opened 2021-06-24
KDEV-1776: Expose MFAService from angular-sdk

DennyGD:


packages/angular-sdk/src/index.ts
```diff
@@ -69,5 +70,6 @@ export {
   LoginOptions,
   MFAContext,
   MFACompleteResult,
-  MFA
+  MFA,
```
ventsislav-georgiev, line 73: Is this export still necessary?

DennyGD: I don't think so. A leftover, probably

ventsislav-georgiev (approved):
LGTM

PR #612, opened 2021-07-01
Fix {N} MFA tests

tsvetomir-nedyalkov:
1. Refactored 'createVerifiedAuthenticator' and 'removeAuthenticator' to use the sdk as they hadn't worked in {N}

2. Fixed totp.js "Buffer is not defined" error by adding a hack with a polyfill for 'buffer'.


DennyGD (approved):
LGTM

PR #616, opened 2021-07-22
KDEV-1814: Remove option to set active user on signup

DennyGD:
* Adjust the `signup` method to work with v6 or higher - an auth token is no longer issued when creating a user, so the option to set the active user on signup must be removed.
Most of the changes are just tests adjustments.


alexanderfilipov (approved):
LGTM. Given that we may bring back the active user check before `signup`.

PR #622, opened 2021-08-13
KDEV-1897: React-native storage optimization

alexanderfilipov:
Usage of `AsyncStorage` has been optimized by storing each document as a separate key-value pair, instead of storing the whole collection as one value and parsing it for every operation. This was especially inefficient for operations "by id".
It was also necessary to change the key format from using `.` as separator to using `:` instead, because of the tagged collections functionality, which uses `.` as part of the collection name and this led to collisions.

### DIFF
--- a/packages/react-native-sdk/src/storage/asyncStorage.ts
+++ b/packages/react-native-sdk/src/storage/asyncStorage.ts
@@ -1,12 +1,14 @@
 import AsyncStorage from '@react-native-async-storage/async-storage';
 
+const SEPARATOR = ':';
+
 function getKey(dbName: string, tableName: string, id: string) {
-  return `${dbName}.${tableName}.${id}`;
+  return `${dbName}${SEPARATOR}${tableName}${SEPARATOR}${id}`;
 }
 
 async function getTableKeys(dbName: string, tableName: string) {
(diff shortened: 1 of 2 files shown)


packages/react-native-sdk/src/storage/asyncStorage.ts
```diff
@@ -1,58 +1,55 @@
...
+  await AsyncStorage.multiSet(kvPairs);
+  return kvPairs;
 }
 
 export async function removeById(dbName: string, tableName: string, id: string) {
-  const table = await getTable(dbName, tableName);
-  if (table.delete(id)) {
-    setTable(dbName, tableName, table);
+  const key = getKey(dbName, tableName, id);
+  const doc = await AsyncStorage.getItem(key);
```
ventsislav-georgiev, line 37: Do we need to check if it exists before removing it?

ventsislav-georgiev: Can we determine if the remove happened from the AsyncStorage.removeItem result?

alexanderfilipov: It doesn't seem so `removeItem` returns a `Promise<void>`.

ventsislav-georgiev (approved):

PR #638, opened 2022-06-09
KDEV-2482: Clear session on InvalidCredentials error

DennyGD:
The PR clears the active user when an InvalidCredentials error is received and MIC is not involved. Essentially, this code is present in the latest SDK version. What is different is the missing call to the logout endpoint which has no way of succeeding.

### DIFF
--- a/packages/js-sdk/src/http/request.ts
+++ b/packages/js-sdk/src/http/request.ts
@@ -2,7 +2,7 @@ import isString from 'lodash/isString';
 import PQueue from 'p-queue';
 import { Base64 } from 'js-base64';
 import { InvalidCredentialsError } from '../errors/invalidCredentials';
-import { getAppSecret} from '../kinvey';
+import { getAppSecret, getAppKey } from '../kinvey';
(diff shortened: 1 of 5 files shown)


packages/js-sdk/src/http/request.ts
```diff
@@ -208,6 +209,20 @@ export class KinveyHttpRequest extends HttpRequest {
                 logger.error(error.message);
               }
             }
+          } else {
```
mkapostolov, line 212: I think we should remove this section as it is only executed in case the login is not with MIC.
This means that the behavior will differ in the case of MIC and not MIC:
- if you are logged with MIC - the session is not automatically cleared
- if you are not logged with MIC - the session is automatically cleared

It would be better if the session is cleared in both cases if the MIC refresh fails for some reason.

DennyGD: Yes, I may have misinterpreted the task's description but it makes sense to clear it in both cases.

mkapostolov (approved):

PR #641, opened 2022-07-12
Fix refresh token overwrite

mkapostolov:
Calling _me endpoint returns the last issued refresh_token as a response,
which might be issued to another device.
Therefore, we should not persist the returned refresh_token from_me endpoint locally on the device.


ivanovit (approved):
LGTM

PR #646, opened 2022-11-29
KDEV-2693: Clear active user on InvalidCredentials

DennyGD:
* The PR adds cleaning up the active user and stores when an InvalidCredentials error occurs and a valid access token cannot be acquired. It intends to bring [this improvement](https://github.com/Kinvey/js-sdk/pull/638) to the 5.x.x version of the SDK. The code is mostly the same but there are some slight differences here and there.

**Note:** I've created a branch from the 5.1.4 tag.


ivanovit (approved):
LGTM

PR #650, opened 2023-06-19
CHAT-4575: Update angular SDK

DennyGD:
The main goal of this PR is to make the angular SDK compatible with angular 12. It brings the following changes:
* updating angular deps from version 7/8 to 12
* updating typescript and rxjs [(compatibility matrix)](https://angular.io/guide/versions)
* tsconfig changes for the js SDK, HTML5 SDK and angular SDK, so that they now target ES6
* updating the angular Kinvey module [(Module with providers)](https://angular.io/guide/migration-module-with-providers)
* replacing lodash with lodash-es (in order to not use common js modules anymore)
* changing the nodejs SDK integration tests to not use the local version of the js SDK (otherwise, the node SDK would have to be updated as well)
* transforming the js SDK unit tests with babel because jest doesn't work with ES6 
* skipping some of the js SDK tests due to errors (TypeError: Cannot set property methodFactory of #<Object> which has only a getter) - this is related to the log module in the js SDK and the way the loglevel-plugin-prefix module is used along with the babel transformation

Angular will still give a warning when the Kinvey angualr SDK is loaded because the url package that is used in the js SDK does not have an ES6 version, it is still exported as common js module. The warning can be skipped, though: [Configuring Common js dependencies](https://angular.io/guide/build#configuring-commonjs-dependencies)

**Note:** Before merging, the SDKs versions should be updated: 
* the version of the js SDK should just be a major version (and not alpha); 
* the version of the js SDK in the package.json of the HTML5 SDK must be changed accordingly plus the version of the HTML5 SDK itself;
* the version of the HTML5 SDK in the package.json of the angular SDK must be changed accordingly plus the version of the angualr SDK itself

During development I've used node v12.16.1 (npm v6.14.15) [(due to compatibility)](https://angular.io/guide/versions).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,6 +3,16 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
+    "@ampproject/remapping": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.1.tgz",
+      "integrity": "sha512-lFMjJTrFL3j7L9yBxwYfCq2k6qqwHyzuUl/XBnif78PWTJYyL/dfowQHWE3sp6U6ZzqWiiIZnpTMO96zhkjwtg==",
+      "dev": true,
+      "requires": {
+        "@jridgewell/gen-mapping": "^0.3.0",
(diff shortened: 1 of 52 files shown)


ivanovit (approved):
LGTM

